PR #11, opened 2022-11-02
Fix for new build

samansmink:
This seems to fix the issue I ran into where when building the extension from the DuckDB CI, it would fail.
@pdet

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,22 +1,28 @@
 cmake_minimum_required(VERSION 2.8.12)
-set(TARGET_NAME substrait)
-project(${TARGET_NAME})
 
-# set(EXTENSION_CLASS ${TARGET_NAME}Extension)
-string(SUBSTRING ${TARGET_NAME} 0 1 FIRST_LETTER)
-string(TOUPPER ${FIRST_LETTER} FIRST_LETTER)
-string(REGEX REPLACE "^.(.*)" "${FIRST_LETTER}\\1" EXTENSION_CLASS
-                     "${TARGET_NAME}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,22 +1,28 @@
...
-project(${TARGET_NAME})
 
-# set(EXTENSION_CLASS ${TARGET_NAME}Extension)
-string(SUBSTRING ${TARGET_NAME} 0 1 FIRST_LETTER)
-string(TOUPPER ${FIRST_LETTER} FIRST_LETTER)
-string(REGEX REPLACE "^.(.*)" "${FIRST_LETTER}\\1" EXTENSION_CLASS
-                     "${TARGET_NAME}")
+if(DEFINED EXTERNAL_EXTENSION_DIRECTORIES)
+    # Extension is being built from DuckDB CMake file, just add the extension dir
+    add_subdirectory(arrow)
```
pdet, line 5: this should be substrait, no?

pdet (changes requested):
Hey Sam, just a small comment.

Thanks for fixing this!

CMakeLists.txt
samansmink: yea that probably works a lot better

samansmink:
I'll close this one for now as there's another issue with this that needs a bit more thought: duckdb currently looks for the extensions only in the root build dir of the extension. Since the new build system moves the extensions one folder down into a folder, currently duckdb scripts/ci do not find it.

we need to fix where out of tree extensions are expected to produce their libraries for the duckdb ci to find them.

PR #13, opened 2022-11-09
fix for binary directory

samansmink:
Same solution as used for the arrow extension @pdet


Mytherin:
Thanks for the quick fix! Let me retry this

PR #15, opened 2022-11-16
Bugfix/readme fix for SQL and python examples

prmoore77:
While working through the SQL and Python examples in the readme, I noticed some issues.  I've attempted to correct these for the next user so that they can hit the ground running.  Please let me know if I can improve this PR.

Thanks!


pdet:
Hi @prmoore77 , thanks for the fixes, it's always better when someone actually tries the readme since they are not as biased as the writer :-)

Issue #10, opened 2022-11-01
Consume substrait from Json - String representation


pdet:
This is necessary to consume plans produced by `Isthmus`

PR #16, opened 2022-12-01
Adding support to read json plans

pdet:
Fix: #10 

CC: @richtia 
After CI Passes, I'll merge it, but then I still have to add a PR at DuckDB to support the latest version of substrait and to have functions in the Python and R API :-)


richtia:
> CC: @richtia
> After CI Passes, I'll merge it, but then I still have to add a PR at DuckDB to support the latest version of substrait and to have functions in the Python and R API :-)

Awesome! Thanks @pdet !

Issue #5, opened 2022-09-13
Fix Mapping for function names in substrait plans.


richtia:
It looks like duckdb doesn't have a mapping between the duckdb SQL function names to substrait function names when creating the substrait plans.  For the most part, the substrait plans function names are the same as the duckdb SQL function name.

Here's a list of functions that should have their names remapped when producing the substrait plans.  Expected substrait function name is in parenthesis.

- [ ] mod (modulus)
- [ ] sum (sum) - the function name in the substrait plan is sum_no_overflow
- [ ] abs (abs) - there don't seem to be any substrait functions generated in the plan for this function
- [ ] count (count) - the function name in the substrait plan is count_star
- [ ] stddev (std_dev)
- [ ] extract (extract) - extract using year `SELECT COL, extract('year' FROM COL)`.  the function name in the substrait plan is year
- [ ] like (like) - `SELECT N_NAME FROM nation.parquet WHERE N_NAME LIKE 'ALGERIA';` uses `and`, `equal`, and `is_not_null` substrait functions in the produced plan
- [ ] prefix (starts_with)
- [ ] suffix (ends_with)
- [ ] substr (substring)
- [ ] length (char_length)
- [ ] isnan (is_nan)
- [ ] isfinite (is_finite)
- [ ] isinf (is_infinite)

LindsayWray:
> @LindsayWray Are we missing any of these? Is this good to close?

Yes, this can be closed. Only missing extract, but I've made a separate issue for that one #24

PR #22, opened 2023-01-18
Remap function names

LindsayWray:
Partially addresses issue #5 

Due to occasional differing function names between DuckDB and Substrait, remapping is necessary. This pr adds unordered maps to remap these function names both ways. These maps will likely need to be expanded in the future.


substrait/from_substrait.cpp
```diff
@@ -21,6 +21,27 @@
 #include "google/protobuf/util/json_util.h"
 
 namespace duckdb {
+const std::unordered_map<std::string, std::string> SubstraitToDuckDB::function_names_remap = {
```
pdet, line 24: Should we have a constexpr for these maps?

substrait/from_substrait.cpp
```diff
@@ -21,6 +21,27 @@
...
+  {"modulus", "mod"},
+  {"std_dev", "stddev"},
+  {"starts_with", "prefix"},
+  {"ends_with", "suffix"}, 
+  {"substring", "substr"},
+  {"char_length", "length"},
+  {"is_nan", "isnan"},
+  {"is_finite", "isfinite"},
+  {"is_infinite", "isinf"},
+  {"like", "~~"}
```
pdet, line 34: Is `like` `~~` being tested?

pdet:
Hey Lindsay, many thanks for the PR! It looks great, I only have some small comments.

It also seems you also have to update the tests that match the binary results, since the binaries have now changed :-)

pdet:
Also, is this PR the one that refers to the issue raised on the discord channel?

substrait/from_substrait.cpp
LindsayWray: I just added an extra test for like!

substrait/from_substrait.cpp
LindsayWray: I looked into it, but unfortunately there is no constexpr constructor support for std::map

LindsayWray:
> Also, is this PR the one that refers to the issue raised on the discord channel?

Yes, I hope this PR resolves that issue

pdet:
The CI is hopefully fixed now :-)

pdet:
LGTM!

PR #23, opened 2023-01-18
Enable optimizer

LindsayWray:
Builds further upon PR #22

When creating a DuckDB query plan certain optimizations are implemented that can impact the translation to a substrait plan.

For example the LIKE operator can be substituted by operations/functions such as: and, equal, is_not_null, ends_with, etc. So in translation to a Substrait query plan the LIKE operator from the initial query gets lost. 
Or when a dataset has no negative values, the abs function can be discarded from the plan. However, if the consequential Substrait query plan would be applied to a differing dataset (containing negative values), the abs function is now optimized out of the plan.

This PR adds the option to build a Substrait plan from an unoptimized DuckDB plan in case these substitutions/ optimizations are not desirable. 

Instead of 
``` 
CALL get_substrait('select * from tbl where v LIKE '%x'');
```
you can add a second boolean argument to disable the optimizer
```
CALL get_substrait('select * from tbl where v LIKE '%x'', false);
```


pdet:
Hey Lindsay, thanks for the PR, it looks great!

Just three comments,

1) Should we force users to utilize the parameter name?
as in 
` CALL get_substrait('select * from tbl where v LIKE '%x'', enable_optimizer=false); `

2) Can we also add it to the get_substrait_json?

3) Can we also add it to the respective Python and R APIs, and include tests?

LindsayWray:
> I will work on it!

LindsayWray:
> Hey Lindsay, thanks for the PR, it looks great!
> 
> Just three comments,
> 
> 1. Should we force users to utilize the parameter name?
>    as in
>    `CALL get_substrait('select * from tbl where v LIKE '%x'', enable_optimizer=false);`
> 2. Can we also add it to the get_substrait_json?
> 3. Can we also add it to the respective Python and R APIs, and include tests?

I will work on it!

Makefile
```diff
@@ -62,7 +62,7 @@ release_r: CLIENT_FLAGS=-DBUILD_R=1
 release_r: release
 
 release_python: CLIENT_FLAGS=-DBUILD_PYTHON=1 -DBUILD_JSON_EXTENSION=1 -DBUILD_FTS_EXTENSION=1 -DBUILD_TPCH_EXTENSION=1 -DBUILD_VISUALIZER_EXTENSION=1 -DBUILD_TPCDS_EXTENSION=1
-release_python: debug
+release_python: release
```
pdet, line 65: Slightly unrelated, but could you also push these to the https://github.com/duckdb/extension-template makefile? :-)

test/python/test_enable_optimizer.py
```diff
@@ -0,0 +1,87 @@
+import pandas as pd
+import duckdb
+
+# def test_optimizer_defaults_to_true(require):
```
pdet, line 4: just to confirm it, the tests are commented out because we first need to merge these to make the tests pass right?

Should we add expect to fail for now or skip them (@pytest.mark.skip or xfail) , instead of commenting them out?

Makefile
LindsayWray: Sure, no problem!

test/python/test_enable_optimizer.py
LindsayWray: Yes that's right! I've altered them for now with @pytest.mark.skip

pdet:
Thanks

PR #40, opened 2023-02-22
Re-enable optimizer tests + other misc chores

Tishj:
This PR enables tests that were previously commented out

Also added support for having the `duckdb` folder live outside of the root folder (through `DUCKDB_DIR` env var)
And fixed + ran `make format`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,13 @@
 
 all: release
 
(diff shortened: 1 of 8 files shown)


pdet:
LGTM thks!

PR #54, opened 2023-05-04
[Dev] Makefile rework

Tishj:
Use the main duckdb Makefile from within our Makefile, this reduces the code duplication within our Makefile drastically.

Note: this is still failing because this PR relies on the changes made here: <https://github.com/duckdb/duckdb/pull/7363>


pdet:
@Tishj could you double-check the status of this?

Tishj:
the R dependency installs are super flakey
`ERROR: dependencies ‘callr’, ‘processx’, ‘waldo’ are not available for package ‘testthat’`

Yet the CI step still completes, that's not desired.
But they shouldn't fail to begin with

I wonder if it's a mirror issue or maybe we want to add a step that clears all the site libraries first, so we do a clean install
in case something is possibly cached still

PR #61, opened 2023-07-28
Update duckdb

samansmink:
This PR updates DuckDB to latest version.

There are a few tests that are broken, I have uncommented the failing tests and added FIXME comments. 
I'll open an issue regarding the tests.


samansmink:
Also switched the R job from MacOS to linux as there appeared to be some upstream issue with the testthat package that would sometimes fail

PR #63, opened 2023-08-11
Fix for query result checking

pdet:


carlopi:
I haven't checked and just opened a duplicate by mistake... https://github.com/duckdblabs/substrait/pull/64

src/substrait_extension.cpp
```diff
@@ -71,13 +71,26 @@ static void VerifySubstraitRoundtrip(unique_ptr<LogicalOperator> &query_plan, Co
...
-	if (!actual_result->Equals(*substrait_result)) {
+	unique_ptr<MaterializedQueryResult> substrait_materialized;
+
+	if (substrait_result->type == QueryResultType::STREAM_RESULT) {
+		auto &stream_query = substrait_result->Cast<duckdb::StreamQueryResult>();
+
+		substrait_materialized = stream_query.Materialize();
+	} else if (substrait_result->type == QueryResultType::MATERIALIZED_RESULT) {
+		substrait_materialized = unique_ptr_cast<QueryResult, MaterializedQueryResult>(std::move(substrait_result));
+	}
```
Tishj, line 86: I was gonna say `else { throw InternalException }` but that will happen when we use `substrait_materialized` and it's nullptr anyways

Tishj (approved):
Thanks, this solution makes sense

I was wondering if it might be better to add a method `Materialize` to QueryResult that encapsulates this, but I'm not sure if that would even work (what would the result be on a MaterializedQueryResult?)

PR #74, opened 2024-02-09
[Dev] Add duckdb's patches

carlopi:


carlopi:
I think PR makes sense (absorbing the patches), unsure how to rescue CI though...

pdet:
Absolutely, I'll have a look in the CI, thanks so much for taking the initiative Carlo!

PR #94, opened 2024-07-22
Updated IBIS requirements

nickgerrets:


gforsyth:
just a heads up that there's a bug in ibis-substrait 4 w.r.t. field index offsets in grouping sets following a join.  There's a fix up in https://github.com/ibis-project/ibis-substrait/pull/1078

pdet:
I'm merging it, but we should have a follow up PR for the connection errors, since they seemed to be caused by us.

PR #96, opened 2024-07-24
Patches and disable MATERIALIZED_CTE optimization

pdet:
Depends on: https://github.com/duckdb/duckdb/pull/13146


pdet:
@nickgerrets I merged this, please notice that I had to skip one of the ibis tests because it was failing.
It should unblock you to continue working on the ibis update.

PR #106, opened 2024-09-09
TPC-DS queries and BETWEEN comparison

nickgerrets:
This PR adds tests for all the TPC-DS queries, skipping those that are failing for the moment.

Also implemented is the BETWEEN comparison operator for DuckDB to Substrait, used a lot in the TPC-DS queries.


test/sql/test_between.test
```diff
@@ -0,0 +1,17 @@
...
+PRAGMA enable_verification
+
+statement ok
+PRAGMA disable_optimizer
+
+statement ok
+create table t as select * from range(100) as t(x)
+
+statement ok
+CALL get_substrait('select * from t where x BETWEEN 4 AND 6');
```
pdet, line 17: I think you need to set `enable_optimizer = false` in the `get_substrait` call.

test/sql/test_substrait_tpcds.test
```diff
@@ -0,0 +1,409 @@
...
+
+statement ok
+PRAGMA enable_verification
+
+statement ok
+CALL dsdgen(sf=0.1)
+
+#Q 1 (DELIM_JOIN)
+#statement ok
+#CALL get_substrait('WITH customer_total_return AS (SELECT sr_customer_sk AS ctr_customer_sk, sr_store_sk AS ctr_store_sk, Sum(sr_return_amt) AS ctr_total_return FROM store_returns, date_dim WHERE sr_returned_date_sk = d_date_sk AND d_year = 2001 GROUP BY sr_customer_sk, sr_store_sk) SELECT c_customer_id FROM customer_total_return ctr1, store, customer WHERE ctr1.ctr_total_return > (SELECT Avg(ctr_total_return) * 1.2 FROM customer_total_return ctr2 WHERE ctr1.ctr_store_sk = ctr2.ctr_store_sk) AND s_store_sk = ctr1.ctr_store_sk AND s_state = ''TN'' AND ctr1.ctr_customer_sk = c_customer_sk ORDER BY c_customer_id LIMIT 100; ')
```
pdet, line 17: Instead of commenting out a query, could we have statement error and the error message it generates?

something like:

```
statement error
CALL get_substrait('WITH customer_total_return AS (SELECT sr_customer_sk AS ctr_customer_sk, sr_store_sk AS ctr_store_sk, Sum(sr_return_amt) AS ctr_total_return FROM store_returns, date_dim WHERE sr_returned_date_sk = d_date_sk AND d_year = 2001 GROUP BY sr_customer_sk, sr_store_sk) SELECT c_customer_id FROM customer_total_return ctr1, store, customer WHERE ctr1.ctr_total_return > (SELECT Avg(ctr_total_return) * 1.2 FROM customer_total_return ctr2 WHERE ctr1.ctr_store_sk = ctr2.ctr_store_sk) AND s_store_sk = ctr1.ctr_store_sk AND s_state = ''TN'' AND ctr1.ctr_customer_sk = c_customer_sk ORDER BY c_customer_id LIMIT 100; ')
----
Not Implemented Exception: DELIM JOIN
```

That won't work if the query crashes, if any do we can leave it commented out for now!

pdet (changes requested):
Thanks for the work Nick!!

I left two minor-ish comments!

test/sql/test_substrait_tpcds.test
```diff
@@ -0,0 +1,496 @@
...
+CALL dsdgen(sf=0.1)
+
+#Q 1 (DELIM_JOIN)
+statement error
+CALL get_substrait('WITH customer_total_return AS (SELECT sr_customer_sk AS ctr_customer_sk, sr_store_sk AS ctr_store_sk, Sum(sr_return_amt) AS ctr_total_return FROM store_returns, date_dim WHERE sr_returned_date_sk = d_date_sk AND d_year = 2001 GROUP BY sr_customer_sk, sr_store_sk) SELECT c_customer_id FROM customer_total_return ctr1, store, customer WHERE ctr1.ctr_total_return > (SELECT Avg(ctr_total_return) * 1.2 FROM customer_total_return ctr2 WHERE ctr1.ctr_store_sk = ctr2.ctr_store_sk) AND s_store_sk = ctr1.ctr_store_sk AND s_state = ''TN'' AND ctr1.ctr_customer_sk = c_customer_sk ORDER BY c_customer_id LIMIT 100; ')
+----
+Not implemented Error: DELIM_JOIN
+
+#Q 2 (results dont match)
+#statement ok
```
pdet, line 22: can this be a statement error?

pdet (changes requested):
Hi @nickgerrets, is it possible to make the remaining failing tests also statement errors with the error messages, instead of commenting them out?

pdet (approved):
LGTM! Thanks Nick! :)

PR #110, opened 2024-09-11
Actually deploy to nightly-extensions.duckdb.org

carlopi:


pdet:
Thanks @carlopi! Works great now!!

PR #112, opened 2024-09-23
Fixing substrait CI

carlopi:
Fixing after extension-ci-tools has changed interface, plus adding tpcds as build/test time dependency and fixing test that had failed on main (https://github.com/duckdb/duckdb/actions/runs/10962094612/job/30441574363?pr=14055)


pdet:
Thanks!!

PR #114, opened 2024-10-29
feat: add explain TableFunction for substrait

drin:
This adds a new table function, `explain_substrait`, as well as a new execution kernel for the function and binding kernel. This reuses the `FromSubstraitFunctionData` structure since it contains all the pieces we would need.

The reason this is useful is because `explain` naively returns the table function itself instead of the plan it may execute. In the case of executing substrait (e.g. `from_substrait(...)`), this would be a single operator for the `from_substrait` function itself. If, instead one calls `explain_substrait(...)`, the result would be an explain plan for what the substrait extension translates into duckdb (the result of [SubstraitPlanToDuckDBRel](https://github.com/duckdb/substrait/blob/main/src/substrait_extension.cpp#L235))

### DIFF
--- a/src/substrait_extension.cpp
+++ b/src/substrait_extension.cpp
@@ -250,10 +250,43 @@ static unique_ptr<FunctionData> FromSubstraitBindJSON(ClientContext &context, Ta
 	return SubstraitBind(context, input, return_types, names, true);
 }
 
+static unique_ptr<FunctionData> ExplainSubstraitBind(ClientContext &context, TableFunctionBindInput &input,
+                                                     vector<LogicalType> &return_types, vector<string> &names) {
+	if (input.inputs[0].IsNull()) {
+		throw BinderException("explain_substrait cannot be called with a NULL parameter");
+	}
+
+	constexpr bool is_json = false;
+	string serialized = input.inputs[0].GetValueUnsafe<string>();
+
+	// We can reuse the `FromSubstraitFunctionData` struct to return the explain plan
+	auto result = make_uniq<FromSubstraitFunctionData>();
+	result->conn = make_uniq<Connection>(*context.db);
+	result->plan = SubstraitPlanToDuckDBRel(*result->conn, serialized, is_json);
+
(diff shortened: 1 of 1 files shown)


drin:
I realized this functionality would be generally useful. For now, this is mostly a placeholder until I can maybe test this at the end of the week and write some tests, etc.

pdet:
Hey Aldrin, thanks for the PR, I think it looks good!

Could you just add a few tests?

drin:
99752e4 is the rebase, and 5707b53 fixes a few issues and adds a test.

Unfortunately, I am developing with protobuf v28.3 and modified my duckdb to link against [mohair-substrait][repo-mohairsubstrait]: so, it is difficult for me to actually test the changes.

at the moment, I'm getting the following error when trying to run against a mostly accurate version of this PR:
```bash
 symbol not found in flat namespace '__ZN6google8protobuf8internal26fixed_address_empty_stringE'
```

I can continue trying to tie this off in a few days once I reach a good checkpoint with other work.


[repo-mohairsubstrait]: https://github.com/drin/mohair-substrait/tree/feat-skytether-engine

drin:
@pdet I was wondering if `ExplainRelation::GetTableRef` would return a `TableRef` that would eventually resolve to an equivalent `QueryResult` as `Relation::Explain` does?

In other words, I am currently returning a `QueryResult` via [substrait_extension.cpp#L312][code-execexplain] and using a normal `TableFunction` (uses a normal bind function). I'm wondering if I could construct an `ExplainRelation` similarly to how the root `ProjectionRelation` is constructed in [TransformRootOp][code-makerel], and then I can call `GetTableRef` on that as you do for the other substrait functions (e.g. `from_substrait`).

Thanks!


[code-execexplain]: https://github.com/drin/duckdb-substrait/blob/feat-explain/src/substrait_extension.cpp#L312
[code-makerel]:       https://github.com/drin/duckdb-substrait/blob/feat-explain/src/from_substrait.cpp#L781

PR #124, opened 2024-11-14
feat: Add support for to_substrait for virtual table expression

anshuldata:


test/c/test_substrait_c_api.cpp
```diff
@@ -45,3 +46,23 @@ TEST_CASE("Test C Get and To Json-Substrait API", "[substrait-api]") {
...
 }
+
+
+TEST_CASE("Test C VirtualTable input Literal", "[substrait-api]") {
+  DuckDB db(nullptr);
+  Connection con(db);
+
+  auto json = con.GetSubstraitJSON("select * from (values (1, 2),(3, 4))");
+  REQUIRE(!json.empty());
+  std::cout << json << std::endl;
```
EpsilonPrime, line 57: Do we want to test anything else besides that there wasn't an error during parsing?  Perhaps call GetSubstrait and check a few things like the number of rows/columns is right, that the type of the read operation is a virtual table, etc.

anshuldata: It will done as part of from_substrait PR. Waiting on a change from duckdb (Constructor having RelationContextWrapper and ParsedExpression) once it is in, same test case will be enhanced to check roundtrip

anshuldata:
included this change as part of #130 . Closing this PR

PR #125, opened 2024-11-18
Support CTAS

scgkiran:


src/from_substrait.cpp
```diff
@@ -698,6 +699,29 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformSetOp(const substrait::Rel &sop
...
+		string schema_name;
+		if (table_idx > 0) {
+			schema_name = nobj.names(0);
+		}
+
+		auto input = TransformOp(swrite.input());
+		return input->CreateRel(schema_name, table_name);
+        }
+	default:
+		throw NotImplementedException("Unsupported write operation" + to_string(swrite.op()));
```
jacques-n, line 721:
```suggestion
		throw NotImplementedException("Unsupported write operation " + to_string(swrite.op()));
```

src/from_substrait.cpp
```diff
@@ -778,6 +804,19 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformRootOp(const substrait::RelRoot
...
+	if (sop.input().rel_type_case() == substrait::Rel::RelTypeCase::kWrite) {
+		auto write = sop.input().write();
+		switch (write.op()) {
+		case substrait::WriteRel::WriteOp::WriteRel_WriteOp_WRITE_OP_CTAS: {
+			const auto create_table = static_cast<CreateTableRelation *>(child.get());
+			auto proj = make_shared_ptr<ProjectionRelation>(create_table->child, std::move(expressions), aliases);
+			return proj->CreateRel(create_table->schema_name, create_table->table_name);
+		}
+		default:
+			return child;
```
jacques-n, line 816: returning child here seems weird. what's the reasoning?

EpsilonPrime: It'd be useful to document this branch as well with the answer.  I'm guessing we're treating this as a no-op for all non-write operations.

jacques-n: I don't understand why would we wouldn't throw here. It sounds like you're saying that code above is going to throw so we just return something random here. And what's wrong with adding a projection relation?

jacques-n: Let's merge this change and open an issue to follow-up on better clarity around why we have these weird specialized cases outside of individual transform ops.

jacques-n (approved):

jacques-n:
As mentioned, going to merge this or as part of cherry picking from other branch and then reevaluate in subsequent issue.

PR #135, opened 2025-01-08
fix: project rel to and from substrait to include pass through columns

scgkiran:


src/from_substrait.cpp
```diff
@@ -492,22 +492,59 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformFilterOp(const substrait::Rel &
...
+	const substrait::RelCommon* common = nullptr;
+	switch (sop.rel_type_case()) {
+	case substrait::Rel::RelTypeCase::kJoin:
+		common = &sop.join().common();
+		break;
+	case substrait::Rel::RelTypeCase::kProject:
+		common = &sop.project().common();
+		break;
+	default:
+		throw InternalException("Unsupported relation type " + to_string(sop.rel_type_case()));
```
EpsilonPrime, line 505: Almost every relation should be in this list.  I'd consider calling anything missing not yet implemented.

src/from_substrait.cpp
```diff
@@ -492,22 +492,59 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformFilterOp(const substrait::Rel &
...
+	}
+	if (!common->has_emit()) {
```
EpsilonPrime, line 507: It may be useful to break this into code to get the common table and code to get the emit from the mapping.  That way you can use .the common structure to update direct and emit if necessary.  You might be able to use templates to get the RelCommon which could reduce the overall amount of code.

src/from_substrait.cpp
```diff
@@ -492,22 +492,59 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformFilterOp(const substrait::Rel &
...
 	RootNameIterator iterator(names);
 
-	for (auto &sexpr : sop.project().expressions()) {
-		expressions.push_back(TransformExpr(sexpr, &iterator));
+	auto input_rel = TransformOp(sop.project().input());
+
+	auto mapping = GetOutputMapping(sop);
+	auto num_input_columns = input_rel->Columns().size();
+	if (mapping.empty()) {
+		for (int i = 1; i <= num_input_columns; i++) {
```
EpsilonPrime, line 525: The first column is numbered at zero.  I found https://substrait.io/tutorial/sql_to_substrait/#field-indices to be useful (in addition to the individual relation pages).

scgkiran: The index `i` is used to create the duckdb column reference **not** substrait. For reference please see [here](https://github.com/substrait-io/duckdb-substrait-extension/blob/acbd4ab61f0ff6756551f73dcafe3ab7f2394dc6/src/from_substrait.cpp#L181-L181).

src/from_substrait.cpp
scgkiran: Ah..! Thanks for pointing out, I missed this.

This also reminds me that I could avoid a project by using output mapping, whenever I only have to change the column order of a give relation.

scgkiran: fixed it

test/c/test_projection.cpp
```diff
@@ -0,0 +1,336 @@
...
+	auto expected_json_str = R"({"relations":[{"root":{"input":{"read":{"baseSchema":{"names":["employee_id","name","department_id","salary"],"struct":{"types":[{"i32":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_NULLABLE"}},{"i32":{"nullability":"NULLABILITY_NULLABLE"}},{"decimal":{"scale":2,"precision":10,"nullability":"NULLABILITY_NULLABLE"}}],"nullability":"NULLABILITY_REQUIRED"}},"projection":{"select":{"structItems":[{},{"field":1},{"field":2},{"field":3}]},"maintainSingularStruct":true},"namedTable":{"names":["employees"]}}},"names":["employee_id","name","department_id","salary"]}}],"version":{"minorNumber":53,"producer":"DuckDB"}})";
+	REQUIRE(json_str == expected_json_str);
+	auto result = FromSubstraitJSON(con, json_str);
+	REQUIRE(CHECK_COLUMN(result, 0, {1, 2, 3, 4, 5}));
+	REQUIRE(CHECK_COLUMN(result, 1, {"John Doe", "Jane Smith", "Alice Johnson", "Bob Brown", "Charlie Black"}));
+	REQUIRE(CHECK_COLUMN(result, 2, {1, 2, 1, 3, 2}));
+	REQUIRE(CHECK_COLUMN(result, 3, {120000, 80000, 50000, 95000, 60000}));
+}
+
+TEST_CASE("Test C Project two passthrough columns with Substrait API", "[substrait-api]") {
```
EpsilonPrime, line 74: This appears to be a ProjectRel test case when in fact it's a test of the ReadRel projection clause.  Can we alter the naming here (and elsewhere) to make that distinction more clear?

test/c/test_projection.cpp
```diff
@@ -0,0 +1,336 @@
...
+	REQUIRE(CHECK_COLUMN(result, 0, {10, 20}));
+}
+
+TEST_CASE("Test C Project 1 input column 1 transformation with Substrait API", "[substrait-api]") {
+	DuckDB db(nullptr);
+	Connection con(db);
+
+	REQUIRE_NO_FAIL(con.Query("CREATE TABLE integers(i INTEGER)"));
+	REQUIRE_NO_FAIL(con.Query("INSERT INTO integers VALUES (10), (20), (30)"));
+	CreateEmployeeTable(con);
```
EpsilonPrime, line 47: Is this table being used here?

test/c/test_projection.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+TEST_CASE("Test C Project on Join with Substrait API", "[substrait-api]") {
+	DuckDB db(nullptr);
+	Connection con(db);
+
+	CreateEmployeeTable(con);
+	CreateDepartmentsTable(con);
+
+	auto result = ExecuteViaSubstraitJSON(con,
+		"SELECT e.employee_id, e.name, d.department_name "
```
EpsilonPrime, line 140: Consider adding an additional column to test that project can add as well as preserve columns.

test/c/test_projection.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+TEST_CASE("Test Project with bad plan", "[substrait-api]") {
+	DuckDB db(nullptr);
+	Connection con(db);
+	con.EnableQueryVerification();
+	REQUIRE_NO_FAIL(con.Query("CREATE TABLE integers(i INTEGER)"));
+	REQUIRE_NO_FAIL(con.Query("INSERT INTO integers VALUES (1), (2), (3), (NULL)"));
+
+	auto query_json =  R"({"relations":[{"root":{"input":{"project":{"input":{"fetch":{"input":{"read":{"baseSchema":{"names":["i"],"struct":{"types":[{"i32":{"nullability":"NULLABILITY_NULLABLE"}}],"nullability":"NULLABILITY_REQUIRED"}},"projection":{"select":{"structItems":[{}]},"maintainSingularStruct":true},"namedTable":{"names":["integers"]}}},"count":"5"}},"expressions":[{"selection":{"directReference":{"structField":{}},"rootReference":{}}}]}},"names":["i"]}}],"version":{"minorNumber":53,"producer":"DuckDB"}})";
```
EpsilonPrime, line 158: Might be useful to document what is wrong so folks don't need to read the entire plan to figure out what it is doing.

test/c/test_projection.cpp
```diff
@@ -0,0 +1,336 @@
...
+TEST_CASE("Test Project with duplicate columns", "[substrait-api]") {
+	DuckDB db(nullptr);
+	Connection con(db);
+	con.EnableQueryVerification();
+	REQUIRE_NO_FAIL(con.Query("CREATE TABLE integers(i INTEGER)"));
+	REQUIRE_NO_FAIL(con.Query("INSERT INTO integers VALUES (1), (2), (3), (NULL)"));
+
+	auto query_json =  R"({"relations":[{"root":{"input":{"project":{"input":{"fetch":{"input":{"read":{"baseSchema":{"names":["i"],"struct":{"types":[{"i32":{"nullability":"NULLABILITY_NULLABLE"}}],"nullability":"NULLABILITY_REQUIRED"}},"projection":{"select":{"structItems":[{}]},"maintainSingularStruct":true},"namedTable":{"names":["integers"]}}},"count":"5"}},"expressions":[{"selection":{"directReference":{"structField":{}},"rootReference":{}}}]}},"names":["i", "integers"]}}],"version":{"minorNumber":53,"producer":"DuckDB"}})";
+	auto res1 = FromSubstraitJSON(con, query_json);
+	REQUIRE(CHECK_COLUMN(res1, 0, {1, 2, 3, Value()}));
```
EpsilonPrime, line 171: Is the result emitting two columns or one?  (It should be two.)

src/to_substrait.cpp
```diff
@@ -853,23 +853,89 @@ substrait::Rel *DuckDBToSubstrait::TransformFilter(LogicalOperator &dop) {
...
+				break;
+			}
+			exp_col_idx++;
+		}
+		if (is_passthrough && child_column_count == exp_col_idx) {
+			// skip the projection
+			return TransformOp(*dop.children[0]);
+		}
+		if (child_column_count == exp_col_idx) {
+			// all input columns are projected, no need for output mapping
```
EpsilonPrime, line 909: If we got a query that output the same column a number of times that matched the input number of columns would it trigger this check incorrectly?

EpsilonPrime (commented):
Having this bug nearly fixed is exciting!

test/sql/test_substrait_tpcds.test
```diff
@@ -138,10 +138,10 @@ statement ok
...
 
 #Q 27 (Reference out of range)
-statement error
-CALL get_substrait('SELECT i_item_id, s_state, Grouping(s_state) g_state, Avg(ss_quantity) agg1, Avg(ss_list_price) agg2, Avg(ss_coupon_amt) agg3, Avg(ss_sales_price) agg4 FROM store_sales, customer_demographics, date_dim, store, item WHERE ss_sold_date_sk = d_date_sk AND ss_item_sk = i_item_sk AND ss_store_sk = s_store_sk AND ss_cdemo_sk = cd_demo_sk AND cd_gender = ''M'' AND cd_marital_status = ''D'' AND cd_education_status = ''College'' AND d_year = 2000 AND s_state IN ( ''TN'', ''TN'', ''TN'', ''TN'', ''TN'', ''TN'' ) GROUP BY rollup ( i_item_id, s_state ) ORDER BY i_item_id, s_state LIMIT 100; ')
-----
-Binder Error: Positional reference 7 out of range (total 6 columns)
+#statement error
```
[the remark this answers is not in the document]
scgkiran: Now it fails with an `Internal error`. I didn't find a way to set the expected error for Internal error. I will look some more.

src/from_substrait.cpp
```diff
@@ -500,22 +500,106 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformFilterOp(const substrait::Rel &
...
+		return &sop.merge_join().common();
+	case substrait::Rel::RelTypeCase::kNestedLoopJoin:
+		return &sop.nested_loop_join().common();
+	case substrait::Rel::RelTypeCase::kWindow:
+		return &sop.window().common();
+	case substrait::Rel::RelTypeCase::kExchange:
+		return &sop.exchange().common();
+	case substrait::Rel::RelTypeCase::kExpand:
+		return &sop.expand().common();
+	default:
```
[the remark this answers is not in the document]
scgkiran: I tried it doesn't work that way. We have to explicitly remember or fix it when we hit this error.

src/to_substrait.cpp
```diff
@@ -853,23 +853,89 @@ substrait::Rel *DuckDBToSubstrait::TransformFilter(LogicalOperator &dop) {
...
 	auto res = new substrait::Rel();
 	auto &dproj = dop.Cast<LogicalProjection>();
+
+	auto child_column_count = GetColumnCount(*dop.children[0]);
+	auto need_output_mapping = true;
+	if (child_column_count <= dproj.expressions.size()) {
+		// check if the projection is just pass through of input columns with no reordering
+		auto exp_col_idx = 0;
+		auto is_passthrough = true;
+		for (auto &dexpr : dproj.expressions) {
```
EpsilonPrime, line 892: Consider putting this logic into a separate function to improve local readability and testing.

EpsilonPrime (approved):
Might want to wait and see if anyone else has some thoughts.

Another test case that might be interesting to have is a project with no new expressions.

src/to_substrait.cpp
scgkiran: done

src/to_substrait.cpp
```diff
@@ -879,37 +879,53 @@ substrait::RelCommon *DuckDBToSubstrait::CreateOutputMapping(vector<int32_t> vec
...
+		auto &dref = dexpr.get()->Cast<BoundReferenceExpression>();
+		if (dref.index != exp_col_idx) {
+			return false;
+		}
+		exp_col_idx++;
+	}
+	return child_column_count == exp_col_idx;
+}
+
+bool DuckDBToSubstrait::NeedsOutputMapping(LogicalProjection &dproj, idx_t child_column_count) {
```
EpsilonPrime, line 901: Is this the same as IsPassthroughProjection?

scgkiran: it is slightly different.

EpsilonPrime: It isn't, there are two lines that are different.

src/include/to_substrait.hpp
```diff
@@ -72,8 +72,13 @@ class DuckDBToSubstrait {
 	substrait::Rel *TransformCreateTable(LogicalOperator &dop);
 	substrait::Rel *TransformInsertTable(LogicalOperator &dop);
 	substrait::Rel *TransformDeleteTable(LogicalOperator &dop);
+	static vector<LogicalType>::size_type GetColumnCount(LogicalOperator &dop);
 	static substrait::Rel *TransformDummyScan();
-	//! Methods to transform different LogicalGet Types (e.g., Table, Parquet)
+	static substrait::RelCommon *CreateOutputMapping(vector<int32_t> vector);
+	static bool IsPassthroughProjection(LogicalProjection &dproj, idx_t child_column_count);
+	static bool NeedsOutputMapping(LogicalProjection &dproj, idx_t child_column_count);
+	//! Methods to transform different LogicalGe:75
```
EpsilonPrime, line 80: accidental change?

EpsilonPrime (dismissed):

src/include/to_substrait.hpp
scgkiran: yes fixed it.

test/sql/test_substrait_tpcds.test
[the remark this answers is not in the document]
scgkiran: changed it to Not Implemented error.

EpsilonPrime (approved):

PR #136, opened 2025-01-10
chore: skip c tests if env var is present

rymurr:
While integrating this extension as an out of tree extension using https://github.com/duckdb/duckdb/tree/main/extension#remote-github-repo and building duckdb I found that the C tests don't compile properly: the submodules aren't checked out in this path. 

This was the simplest and cleanest way I could find to skip the compilation of the C tests when compiling substrait directly in duckdb (these tests aren't run in that path anyways and the tests still are run as part of normal CI)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,7 +107,11 @@ add_library(${EXTENSION_NAME} STATIC ${EXTENSION_SOURCES})
 
 set(PARAMETERS "-warnings")
 build_loadable_extension(${TARGET_NAME} ${PARAMETERS} ${EXTENSION_SOURCES})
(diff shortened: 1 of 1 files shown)


rymurr:
> Seems good for now. Wonder if we should make tests work/run in that path at some point.

100% agree. It appears this is the first/only extension that has C tests so is the first one that needs C headers present. I am sure there is a way to tell cmake to change where it looks for include files or do some symlinking but its beyond my cmake skills so wanted to defer until I am more comfortable w cmake

anshuldata (approved):
lgtm

anshuldata (approved):

PR #140, opened 2025-02-05
feat: handle virtual table with zero columns in from_substrait

scgkiran:


src/from_substrait.cpp
```diff
@@ -700,6 +700,10 @@ shared_ptr<Relation> SubstraitToDuckDB::GetValuesExpression(const google::protob
 		}
 		expressions.emplace_back(std::move(expression_row));
 	}
+	if (expressions.size() == 1 && expressions[0].empty()) {
+		// add a dummy row, as DuckDB does not support zero column ValueRelation
+		expressions[0].emplace_back(make_uniq<ConstantExpression>(Value::BOOLEAN(true)));
```
EpsilonPrime, line 705: Does this match the type of the expression)

src/from_substrait.cpp
```diff
@@ -700,6 +700,10 @@ shared_ptr<Relation> SubstraitToDuckDB::GetValuesExpression(const google::protob
 		}
 		expressions.emplace_back(std::move(expression_row));
 	}
+	if (expressions.size() == 1 && expressions[0].empty()) {
+		// add a dummy row, as DuckDB does not support zero column ValueRelation
```
EpsilonPrime, line 704: Does adding a row change what the result would have been?  For example does a select * now return a row when it wouldn't have before?  This transformation is only safe when the result isn't consumed (no field references by its consumer or theirs until the column disappears).

scgkiran: Sorry the comment was wrong, it should be  `dummy column`.

I am adding this dummy column when there is only one row and no columns in the virtual table. 
Yes it is safe only when there are no column references. If there are any column references then the plan is invalid to start with.

src/from_substrait.cpp
scgkiran: no

test/c/test_substrait_c_api.cpp
```diff
@@ -655,3 +655,23 @@ TEST_CASE_METHOD(DataDirectoryFixture, "Test C Iceberg Substrait Snapshot Timest
...
+	REQUIRE(json_str == expected_json_str);
+	auto result = con.FromSubstraitJSON(json_str);
+	REQUIRE(CHECK_COLUMN(result, 0, {1}));
+}
+
+TEST_CASE("Test C Project on empty virtual table for SELECT 1", "[substrait-api]") {
+	DuckDB db(nullptr);
+	Connection con(db);
+
+	auto json_str = R"({"version":{"minorNumber":29, "producer":"substrait-go"}, "relations":[{"root":{"input":{"project":{"common":{"direct":{}}, "input":{"read":{"common":{"direct":{}}, "baseSchema":{"struct":{"nullability":"NULLABILITY_REQUIRED"}}, "virtualTable":{"expressions":[{}]}}}, "expressions":[{"literal":{"decimal":{"value":"AQ==", "precision":1}, "nullable":true}}]}}, "names":["?column?"]}}]})";
```
EpsilonPrime, line 674: Please also test what happens when a virtual table is presented but the input is consumed.  I think you'll find that the number of output columns and names will be mismatched.

scgkiran: Sorry, didn't follow this. how can something consume when there are no columns in the virtual table? can you explain more.

EpsilonPrime: Columns are passed from one relation to the next.  Project (if it was working correctly) would add its expressions to the columns from its input.  So a project("1", virtualtable(a)) would end up returning two columns ("1", a).  If the names were just "1" then you'd have a mismatch.  Right now project doesn't add its input columns so the problem is less noticeable.

I am somewhat surprised that DuckDB is not complaining about the base schema requesting less columns than were provided.  If the base schema entry was present I'm fairly certain that the error described above would occur (if not now but when project is fixed).  But since it's not doing a check it looks like this workaround will work.  It's worth having a check that the number of columns in the output is just one (if the current check doesn't enforce that) for your test to make sure that we catch when something changes that might break this workaround.

scgkiran: DuckDb project have explicit expressions, nothing is passthrough from input, thats why this workaround will work.

Yes, I have to take care of this in project fix for from_substrait, especially the code that depends on number of input columns.

EpsilonPrime (approved):

PR #143, opened 2025-02-10
Bump submodules to v1.2.0

rymurr:
This does a few things:
1. Replaces the `Connection` members which were removed in duckdb/duckdb#15810
2. bump duckdb and other submodules to v1.2
3. removes python and R tests as they were reliant on the direct connection access in duckdb/duckdb#15810
4. fix substrait bugs that cropped up during upgrade
5. enable a previously failing test


scgkiran (approved):

rymurr:
A point of clarity on removing R and python:

duckdb originally had substrait support directly on the `Connection` object. They also had substrait support in adbc. Since the removal of substrait as a core extension they have removed support in adbc and on the `Connection` object. The primary point of the python and R tests was to test these functions. Since they no longer exist the corresponding tests are superflouous. One can still call substrait functions from R and Python by calling the table functions directly

Issue #145, opened 2025-02-10
chore: Update documentation for 1.2.0


EpsilonPrime:
As part of the Substrait extension moving to be a community extension the [documentation on Substrait has been removed from the DuckDB site](https://github.com/duckdb/duckdb-web/pull/4749).  We should integrate that documentation in this repository and in [the extension's documentation](https://github.com/duckdb/community-extensions/pull/285).

PR #147, opened 2025-02-26
fix: update README for DuckDB 1.2.0

amoeba:
Closes https://github.com/substrait-io/duckdb-substrait-extension/issues/145.

I took the chance here to do a full editing pass on the README so there are more changes than strictly required to update the README for 1.2.0. I'm happy to split this up if that's preferred.

From a high level perspective, I:

- Am proposing a new name for the project. The previous "Substrait - DuckDB" reads like "Substrait Minus DuckDB" and I've noticed other community extensions are adopting a similar naming pattern to what I proposed here.
- Removed the references to the Python and R function bindings since they were removed in 1.2.0. This simplified the document.
- Updated references from "install substrait" to "install substrait from community"
- Reformatted and linted the document

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,65 +1,89 @@
-# Substrait - DuckDB
-Substrait - DuckDB is an extension that provides substrait support to [DuckDB](https://www.duckdb.org).
-The main goal of this extension is to support a both production and consumption of substrait query plans in DuckDB.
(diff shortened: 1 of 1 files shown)


EpsilonPrime (approved):

EpsilonPrime:
Thanks for taking this on!

PR #152, opened 2025-03-07
feat: add sql test workflow

EpsilonPrime:


scgkiran:
I think we already run this on different arch already, like [here](https://github.com/substrait-io/duckdb-substrait-extension/actions/runs/13713520040/job/38354282564?pr=152)

Am I missing something?

EpsilonPrime:
That's what I was looking for, thanks!

PR #153, opened 2025-03-07
feat: Return not implemented errors instead of internal errors

EpsilonPrime:
For unsupported types we now return not implemented errors instead of internal errors.  This means that subsequent queries will be able to proceed instead of requiring DuckDB to be restarted.

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -137,7 +137,7 @@ Value TransformLiteralToValue(const substrait::Expression_Literal &literal) {
 		case PhysicalType::INT128:
 			return Value::DECIMAL(substrait_value, substrait_decimal.precision(), substrait_decimal.scale());
 		default:
-			throw InternalException("Not accepted internal type for decimal");
+			throw NotImplementedException("Unsupported internal type for decimal: " + decimal_type.ToString());
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/from_substrait.cpp
```diff
@@ -174,7 +174,7 @@ Value TransformLiteralToValue(const substrait::Expression_Literal &literal) {
 	case substrait::Expression_Literal::LiteralTypeCase::kVarChar:
 		return {literal.var_char().value()};
 	default:
-		throw SyntaxException("literals of this type number are not implemented: " + to_string(literal.literal_type_case()));
+		throw NotImplementedException("literals of this type number are not implemented: " + to_string(literal.literal_type_case()));
```
scgkiran, line 177: Should we try to get the field name and use it in the exception message?
`literal.GetDescriptor()->FindFieldByNumber(literal.literal_type_case())->name()`

scgkiran (approved):

scgkiran (approved):

src/from_substrait.cpp
EpsilonPrime: Nice, I've updated everywhere I could find to use this construct.

scgkiran (approved):

PR #154, opened 2025-03-11
changes to use 1.2.1 version of duckdb and extension-ci-tools.

scgkiran:


rymurr (approved):

.github/workflows/main_distribution.yml
```diff
@@ -21,8 +21,8 @@ jobs:
     name: Build extension binaries
     uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@main
     with:
-      duckdb_version: c009b4ea62f9fcf6ffed6aad51c8f1833121bbfb
-      ci_tools_version: 8262e6fafa6b62a82f9d671d33d311b9b578127e
+      duckdb_version: 8e52ec43959ab363643d63cb78ee214577111da4
```
EpsilonPrime, line 24: It feels like this ought to be the main version of duckdb based on the name of the workflow but the other one had the main label?

EpsilonPrime (approved):

.github/workflows/main_distribution.yml
rymurr: why do we even have two nearly identical worfklows? Seems like main will occasionally break and (maybe) stop us from merging?

scgkiran: Initially both were pointing to main, then we changed it to specific commit-id in [this ](https://github.com/substrait-io/duckdb-substrait-extension/commit/6ae6557bf9427faba90c7fc5ccff2eb8b422f1e1)commit.

distribution.yml workflow is not triggered as part of this PR.

PR #159, opened 2025-04-10
chore: update repo to match v1.2.2

EpsilonPrime:


EpsilonPrime:
The test failure is due to the Iceberg extension not being published yet for 1.2.2.  It looks like that should happen after https://github.com/duckdb/duckdb-iceberg/pull/142 is merged.

EpsilonPrime:
The version being downloaded from the duckdb CLI is related to https://github.com/duckdb/duckdb-iceberg/pull/193.  We don't have anything referring to that or a way to get that version right now (we probably should download the extension directly and use that local version instead of core for more control).

For now I'm going to disable these tests so we can get the build out.

jacques-n (approved):

PR #160, opened 2025-05-01
add support for nested types

achille-roussel:
This PR adds missing support for nested types.

I ran `make format`, hence some of the formatting changes.


test/sql/test_substrait_nested_types.test
```diff
@@ -0,0 +1,46 @@
...
+# description: Test nested type support in substrait extension (LIST, MAP, ARRAY)
+# group: [sql]
+
+require substrait
+
+statement ok
+PRAGMA enable_verification
+
+statement ok
+CREATE TABLE nested_test (
```
EpsilonPrime, line 11: I'd like to see additional nesting (a struct, list, or map within one of the others).

test/sql/test_substrait_nested_types.test
```diff
@@ -0,0 +1,46 @@
+# name: test/sql/test_substrait_nested_types.test
+# description: Test nested type support in substrait extension (LIST, MAP, ARRAY)
```
EpsilonPrime, line 2: What's the difference between a list and an array?  Is STRUCT supposed to be listed here?

EpsilonPrime (commented):
Thanks for adding this additional support!

test/sql/test_substrait_nested_types.test
achille-roussel: Ah sorry, leftover... I tried to implement the ARRAY type (fixed-length list basically), and reverted it because I didn't have a clear use case for it, and it was harder to test due to Substrait not having fixed-length arrays.

I'll remove ARRAY from the comment here.

achille-roussel:
I went ahead and addressed the requested changes and added more tests, let me know if anything else needs to be changed!

EpsilonPrime (approved):

achille-roussel:
Thanks for the quick turn around and feedback!

PR #169, opened 2025-12-17
feat: update to DuckDB v1.4.3

andrew-coleman:
To get the same tests to pass required implementing
support for the precision_timestamp type, and
support for materialised CTEs.


src/from_substrait.cpp
```diff
@@ -962,6 +988,15 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformWriteOp(const substrait::Rel &s
 	}
 }
 
+shared_ptr<Relation> SubstraitToDuckDB::TransformReferenceOp(const substrait::Rel &sop) {
+	auto index = sop.reference().subtree_ordinal();
+	if (index >= ctes.size()) {
+		throw InvalidInputException("Reference made to non-existent top-level relation: %d\n", index);
```
bestbeforetoday: Existing exception messages don't seem to include a trailing newline.

src/to_substrait.cpp
```diff
@@ -1814,6 +1814,15 @@ substrait::Rel *DuckDBToSubstrait::TransformDeleteTable(LogicalOperator &dop) {
 	return rel;
 }
 
+substrait::Rel *DuckDBToSubstrait::TransformCTERef(LogicalOperator &dop) {
+	auto rel = new substrait::Rel();
+	auto &cte_ref = dop.Cast<LogicalCTERef>();
+	auto index = find(cte_indices.begin(), cte_indices.end(), cte_ref.cte_index) - cte_indices.begin();
```
bestbeforetoday: The `find` will return `cte_indices.end()` if `cte_ref.cte_index` is not found. If this might ever happen then perhaps a guard against that case should be included.

test/sql/test_substrait_parquet.test
```diff
@@ -36,5 +36,5 @@ CALL get_substrait('select * from parquet_scan(''data/parquet-testing/glob*/t?.p
 statement error
 CALL get_substrait('select * from parquet_scan('''') order by i')
 ----
-No files found that match the pattern ""
+Binder Error: Referenced column "i" not found in FROM clause!
```
bestbeforetoday, line 39: Not sure why the error message changed here. Is it just a change in DuckDB behaviour with the updated version?

bestbeforetoday (commented):
To my non-expert eye, looks generally good. A few queries in-line to consider — or ignore if appropriate.

This change does not add any tests that exercise common table expressions. It might be worth adding something simple.

test/sql/test_substrait_parquet.test
andrew-coleman: Yes, now it looks like parquet_scan('') returns an empty table.

andrew-coleman:
> This change does not add any tests that exercise common table expressions. It might be worth adding something simple.

Lots of the tests have CTEs (WITH clause). It's just that earlier versions of DuckDB used to inline them.
https://duckdb.org/2024/09/09/announcing-duckdb-110#automatic-cte-materialization

bestbeforetoday (approved):
Looks good to me!

PR #176, opened 2026-05-06
feat: upgrade substrait to v0.89.0

andrew-coleman:
Update substrait dependency.
Fixup code and tests to handle changes to the substrait protos:
- change of URI to URN
- changed representation of grouping expressions


src/from_substrait.cpp
```diff
@@ -169,7 +169,19 @@ Value TransformLiteralToValue(const substrait::Expression_Literal &literal) {
...
 		interval.days = literal.interval_day_to_second().days();
-		interval.micros = literal.interval_day_to_second().microseconds();
+		// Convert seconds to microseconds and add subseconds (with precision adjustment)
+		int64_t seconds_in_micros = literal.interval_day_to_second().seconds() * Interval::MICROS_PER_SEC;
+		int64_t subseconds = literal.interval_day_to_second().subseconds();
+		int32_t precision = literal.interval_day_to_second().precision();
+		// Adjust subseconds based on precision (e.g., precision 9 = nanoseconds, need to convert to microseconds)
+		if (precision > 6) {
+			// Convert from higher precision (e.g., nanoseconds) to microseconds
+			subseconds = subseconds / (int64_t)pow(10, precision - 6);
```
bestbeforetoday, line 180: Does using the `pow` function mean we should have `#include <cmath>`?

src/to_substrait.cpp
```diff
@@ -596,19 +599,17 @@ uint64_t DuckDBToSubstrait::RegisterFunction(const string &name, vector<::substr
 		throw InternalException("Missing function name");
 	}
 	auto function = custom_functions.Get(name, args_types);
-	auto substrait_extensions = plan.mutable_extension_uris();
+	auto substrait_extensions = plan.mutable_extension_urns();
 	if (!function.IsNative()) {
-		auto extensionURI = function.GetExtensionURI();
-		auto it = extension_uri_map.find(extensionURI);
+		auto extensionURN = function.GetExtensionURI();
```
bestbeforetoday: Should `GetExtensionURI()` also change to `GetExtensionURN()`?

src/to_substrait.cpp
```diff
@@ -596,19 +599,17 @@ uint64_t DuckDBToSubstrait::RegisterFunction(const string &name, vector<::substr
...
+		auto it = extension_uri_map.find(extensionURN);
```
bestbeforetoday: Should `extension_uri_map` change to `extension_urn_map`

src/to_substrait.cpp
```diff
@@ -596,19 +599,17 @@ uint64_t DuckDBToSubstrait::RegisterFunction(const string &name, vector<::substr
...
-			auto allocated_string = new string();
-			*allocated_string = extensionURI;
-			auto uri = new substrait::extensions::SimpleExtensionURI();
-			uri->set_allocated_uri(allocated_string);
-			uri->set_extension_uri_anchor(last_uri_id);
-			substrait_extensions->AddAllocated(uri);
+			extension_uri_map[extensionURN] = last_uri_id;
+			auto urn = new substrait::extensions::SimpleExtensionURN();
+			urn->set_urn(extensionURN);
+			urn->set_extension_urn_anchor(last_uri_id);
```
bestbeforetoday: Should `last_uri_id` change to `last_urn_id`?

test/sql/test_spark_gateway.test
```diff
@@ -14,95 +14,95 @@ call dbgen(sf=0.01)
 
 statement ok
 CALL from_substrait_json('{
-  "extensionUris": [
+  "extensionUrns": [
     {
-      "extensionUriAnchor": 1,
-      "uri": "/functions_comparison.yaml"
+      "extensionUrnAnchor": 1,
+      "urn": "extension:substrait:functions_comparison"
```
bestbeforetoday: The URN format here is quite different from the ones in the other tests touched by this change. I guess since the original URIs are also different means this is expected?

bestbeforetoday (commented):
Generally looks good to me. A few questions in comments.

src/from_substrait.cpp
andrew-coleman: Good question.  I guess it's being included by another header file somewhere.  I'll add it explicitly here.

src/to_substrait.cpp
andrew-coleman: Yes it should...  and the string it generates.  Good catch.

bestbeforetoday (approved):
Looks good to me.

PR #178, opened 2026-05-12
feat: implement grouping sets

andrew-coleman:
Required to support ROLLUP. Fixes a number of TPCDS tests.

This PR was assisted by AI.

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -699,23 +699,34 @@ SubstraitToDuckDB::TransformProjectOp(const substrait::Rel &sop,
 }
 
 shared_ptr<Relation> SubstraitToDuckDB::TransformAggregateOp(const substrait::Rel &sop) {
-	vector<unique_ptr<ParsedExpression>> groups, expressions;
+	vector<unique_ptr<ParsedExpression>> expressions;
+	GroupByNode group_node;
 
 	auto input_rel = TransformOp(sop.aggregate().input());
 
 	if (sop.aggregate().groupings_size() > 0) {
-		for (auto &sgrp : sop.aggregate().groupings()) {
+		// First, add ALL grouping expressions to the GroupByNode
+		for (int i = 0; i < sop.aggregate().grouping_expressions_size(); i++) {
+			auto expr = TransformExpr(sop.aggregate().grouping_expressions(i));
+			// Add to output expressions
+			expressions.push_back(expr->Copy());
+			// Add to group_node
+			group_node.group_expressions.push_back(std::move(expr));
+		}
+		
+		// Now convert each Substrait grouping to a DuckDB grouping set
+		for (int g = 0; g < sop.aggregate().groupings_size(); g++) {
+			auto &sgrp = sop.aggregate().groupings(g);
+			GroupingSet grouping_set;
 			// expression_references contains indices into the grouping_expressions array
 			for (auto ref_idx : sgrp.expression_references()) {
-				// Get the expression from the AggregateRel's grouping_expressions array
 				if (ref_idx < (uint32_t)sop.aggregate().grouping_expressions_size()) {
-					auto expr = TransformExpr(sop.aggregate().grouping_expressions(ref_idx));
-					groups.push_back(expr->Copy());
-					expressions.push_back(std::move(expr));
+					grouping_set.insert(ref_idx);
 				} else {
 					throw InternalException("Invalid expression reference index in grouping");
 				}
 			}
+			group_node.grouping_sets.push_back(grouping_set);
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


src/substrait_extension.cpp
```diff
@@ -180,6 +180,12 @@ static void VerifySubstraitRoundtrip(unique_ptr<LogicalOperator> &query_plan, Cl
 	auto &subs_col_coll = substrait_materialized->Collection();
 	string error_message;
 	if (!ColumnDataCollection::ResultEquals(actual_col_coll, subs_col_coll, error_message)) {
+		// Check if the error is due to NaN vs NULL mismatch (common with division by zero)
+		if (error_message.find("nan") != string::npos && error_message.find("NULL") != string::npos) {
```
bestbeforetoday: This looks quite fragile. When I tried removing it, the error message starts with `nan <> NULL`. If there is no more robust way to check and this is the only expected error, maybe that could be checked explicitly to reduce the chance of ignoring a real error?

```suggestion
		if (error_message.rfind("nan <> NULL", 0) != string::npos) {
```

src/from_substrait.cpp
```diff
@@ -824,8 +835,52 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformReadOp(const substrait::Rel &so
...
+			} else {
+				// Fallback: empty result with no schema
+				vector<vector<Value>> empty_rows;
+				if (acquire_lock) {
+					scan = make_shared_ptr<ValueRelation>(context, empty_rows, column_names);
+				} else {
+					scan = make_shared_ptr<ValueRelation>(context_wrapper, empty_rows, column_names);
+				}
+			}
 		}
```
bestbeforetoday: I had a hard time getting my head around the logical flow here. The results of the for loop within the first if block determine which branch is taken for the `if (!column_types.empty())`.

I tried refactoring slightly so the flow was clearer (to me). What do you think?

```suggestion
		} else if (sget.has_base_schema() && sget.base_schema().names_size() > 0 && sget.base_schema().struct_().types_size() > 0) {
			// Empty virtual table represents an empty result (EMPTY_RESULT operator)
			// Extract schema from base_schema
			vector<string> column_names;
			vector<LogicalType> column_types;
			
			auto &base_schema = sget.base_schema();
			auto &struct_type = base_schema.struct_();
			
			// Extract column names and types from base_schema
			for (int i = 0; i < base_schema.names_size() && i < struct_type.types_size(); i++) {
				column_names.push_back(base_schema.names(i));
				column_types.push_back(SubstraitToDuckType(struct_type.types(i)));
			}
			
			// Create a ValueRelation with one row of NULLs to preserve schema, then filter with FALSE to get 0 rows
			// Using Filter with FALSE ensures the row is eliminated before any aggregations
			vector<vector<Value>> one_null_row;
			vector<Value> null_values;
			for (auto &type : column_types) {
				null_values.push_back(Value(type));
			}
			one_null_row.push_back(null_values);
			
			shared_ptr<Relation> values_rel;
			if (acquire_lock) {
				values_rel = make_shared_ptr<ValueRelation>(context, one_null_row, column_names);
			} else {
				values_rel = make_shared_ptr<ValueRelation>(context_wrapper, one_null_row, column_names);
			}
			// Filter with 1=0 to get empty result with schema (eliminates row before aggregation)
			scan = values_rel->Filter("1=0");
		} else {
			// Fallback: empty result with no schema
			vector<vector<Value>> empty_rows;
			vector<string> column_names;
			if (acquire_lock) {
				scan = make_shared_ptr<ValueRelation>(context, empty_rows, column_names);
			} else {
				scan = make_shared_ptr<ValueRelation>(context_wrapper, empty_rows, column_names);
			}
		}
```

bestbeforetoday (commented):
As far as I can tell, looks OK. A couple of suggestions below to consider where the logical flow in the code was confusing for me, or it looked fragile and potentially error-prone. Up to you if / how you want to change what's already here.

src/substrait_extension.cpp
andrew-coleman: It's a fair point, although the `nan <> NULL` bit is not at the end of the error message, so Ill modify it slightly to:

`if (error_message.find("nan <> NULL") != string::npos) {`

src/from_substrait.cpp
andrew-coleman: Yes, I'm happy with that.

src/substrait_extension.cpp
bestbeforetoday: Discussed offline but mentioned here for others... The rfind with position constraint as the second parameter is making it behave as a "starts with" function since it does not search beyond the starting character position for the substring.

The change can use position 1 instead of 0 to allow for `-nan` or `nan` depending on which architecture the code is compiled and run on.

bestbeforetoday (approved):

PR #181, opened 2026-05-13
feat: support RIGHT SEMI join type

andrew-coleman:
This resolves a number of TPC-DS test failures.

This PR was assisted by AI.


bestbeforetoday (approved):
Converting the right semi to a left semi join blew my mind slightly, but the implementation of that translation looks correct to me.

PR #182, opened 2026-05-13
feat: add support for Window relation

andrew-coleman:
Enables all of the TPC-DS tests that rely on windowing functions to pass.
Also includes comprehensive test coverage of all supported functions and options.

This PR was assisted by AI.

### DIFF
--- a/src/to_substrait.cpp
+++ b/src/to_substrait.cpp
@@ -1442,7 +1442,7 @@ substrait::Rel *DuckDBToSubstrait::TransformWindow(LogicalOperator &dop) {
 	substrait::Rel *current_input = TransformOp(*dop.children[0]);
 	
 	for (auto &spec : window_specs) {
-		auto res = new substrait::Rel();
+		auto res = make_uniq<substrait::Rel>();
 		auto swindow = res->mutable_window();
 		
 		// Set the input (either the original input or the previous window relation)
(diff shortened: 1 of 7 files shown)


src/to_substrait.cpp
```diff
@@ -1376,6 +1376,327 @@ substrait::Rel *DuckDBToSubstrait::TransformAggregateGroup(LogicalOperator &dop)
...
+				}
+			} else {
+				throw InternalException("Window boundary expression missing for FOLLOWING");
+			}
+			break;
+		default:
+			// Default to UNBOUNDED PRECEDING
+			lower_bound->mutable_unbounded();
+			break;
+		}
```
bestbeforetoday: This is so close to being identical to the code below to transform the end bound, it might be worth factoring out to a helper function to avoid the large amount of duplication.

src/to_substrait.cpp
```diff
@@ -1376,6 +1376,327 @@ substrait::Rel *DuckDBToSubstrait::TransformAggregateGroup(LogicalOperator &dop)
...
+			current_spec.expression_indices.push_back(i);
+			window_specs.push_back(std::move(current_spec));
+		}
+	}
+	
+	// Now create chained window relations, one for each unique spec
+	substrait::Rel *current_input = TransformOp(*dop.children[0]);
+	
+	for (auto &spec : window_specs) {
+		auto res = new substrait::Rel();
```
bestbeforetoday: If an exception is thrown during this iteration of the for loop, will this new pointer be left without cleanup and potentially leak memory? Could the code below sit in a try-catch block that deletes `res` the re-throws the exception?

andrew-coleman: good catch - should have used make_uniq

bestbeforetoday (approved):

PR #185, opened 2026-05-19
feat: implement grouping function

andrew-coleman:


src/to_substrait.cpp
```diff
@@ -1373,6 +1370,34 @@ substrait::Rel *DuckDBToSubstrait::TransformAggregateGroup(LogicalOperator &dop)
...
+		for (auto &group_idx : grouping_func) {
+			// Each argument is a field reference to a grouping expression
+			auto s_arg = smeas->add_arguments();
+			auto field_ref = s_arg->mutable_value();
+			CreateFieldRef(field_ref, group_idx);
+			
+			// Get the type of the grouping column
+			if (group_idx < daggr.groups.size()) {
+				args_types.emplace_back(DuckToSubstraitType(daggr.groups[group_idx]->return_type));
+			}
```
bestbeforetoday: I can't figure out if it is safe for the CeateFieldRef (and preceding lines) to be called if the `group_idx < daggr.groups.size()` test fails. Putting the whole body of the for loop within the if test still passes the tests. Removing the if test actually passes all the tests too, but having the bounds check before array access seems like good defensive programming. It is just a question of whether any of the code in the for loop should run if the bounds check fails. Or is it an error condition?

andrew-coleman: I think it's probably an error condition if the grouping index exceeds the bounds of the array.  I've updated the code to throw an exception.

bestbeforetoday (approved):

Issue #188, opened 2026-05-20
from_substrait ignores multi-component namedTable.names, cannot resolve tables in attached databases


thgschwind:
### What happens?

TransformReadOp in from_substrait.cpp only reads names(0) from namedTable.names and looks it up with DEFAULT_SCHEMA:

```
auto table_name = sget.named_table().names(0);
// ...
auto table_info = TableInfo(*context, DEFAULT_SCHEMA, table_name);
This means multi-component names like ["mydb", "suppliers"] or ["mydb", "main", "suppliers"] are misinterpreted — "mydb" is treated as the table name, and the actual table name is discarded.
```

Impact:

Tables in attached databases (via ATTACH) live in a separate catalog namespace and require at minimum a 2-component reference to be resolved. Any substrait producer that emits qualified namedTable.names for attached databases will fail with:

Catalog Error: Table with name mydb does not exist!
Did you mean "mydb.suppliers"?
Expected behavior:

The extension should parse all components of namedTable.names:
1. component ["table"] → current behavior (DEFAULT_SCHEMA)
2. components ["db", "table"] → resolve via attached database
3. components ["catalog", "schema", "table"] → resolve via catalog + schema
Suggested fix:

We have a working patch at https://github.com/thgschwind/duckdb-substrait-extension (commits 0caf18d7086154c7a2c1b928da826232e94673dc and 2640d77111e3b45d409efd453999177e706a45f4) with a self-contained Python test that demonstrates the issue and verifies the fix.

### To Reproduce

Run `test/python/test_named_table_resolution.py` in commit 0caf18d7086154c7a2c1b928da826232e94673dc.  There is also a `README.md` in the same directory.
Running the same test on commit 2640d77111e3b45d409efd453999177e706a45f4 should succeed.

For for introducing python, the C test would have been long, sql only I did not manage.

### OS:

osx

### Substrait-Extension Version:

latest from git

### DuckDB Version:

1.4.3

### DuckDB Client:

python

### Have you tried this on the latest `master` branch?

- [x] I have

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [x] I have

thgschwind:
resolved by PR #189

PR #189, opened 2026-05-20
feat: support multi-component namedTable.names for attached database resolution

thgschwind:
Fixes `from_substrait` to handle 2 and 3-component `namedTable.names`, enabling table resolution in attached databases.

Previously only `names(0)` was read, making `["mydb", "suppliers"]` fail with "Table mydb does not exist". Now the leading component is used as the effective schema, which DuckDB resolves through its attached database search path.

Includes a self-contained Python integration test (`test/python/`).

Resolves issue #188.

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -797,15 +797,33 @@ unique_ptr<TableDescription> TableInfo(ClientContext &context, const string &sch
 	return result;
 }
 
+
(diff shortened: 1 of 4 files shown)


thgschwind:
Commit https://github.com/substrait-io/duckdb-substrait-extension/pull/189/commits/60a5b1fbe161399166be767e92a32d930d054956 simplifies the test case.

thgschwind:
Changed the tests to use one of the standard test harnesses.

andrew-coleman (approved):
Looks good. Thanks!

PR #191, opened 2026-05-27
Fix schema mismatch crash

### LINKED ISSUE #190: SIGSEGV in TransformProjectOp when actual table has fewer columns than plan baseSchema
### What happens?

`from_substrait` crashes with a segfault (SIGSEGV, exit code 139) when executing a plan whose baseSchema declares more columns than the actual table has at runtime. This occurs when the plan was generated against one table schema but executed against a table with the same name but fewer columns (e.g., an attached database with a different schema).

## Root Cause
In `src/from_substrait.cpp`, `TransformProjectOp` computes expression indices using the runtime column count `(input_rel->Columns().size())` but the emit outputMapping was generated based on the plan's `baseSchema` column count:

```
// Line ~695
expressions[i] = TransformExpr(sop.project().expressions(mapping[i] - num_input_columns), &iterator);
```

When the actual table has fewer columns than declared in baseSchema:
1. `num_input_columns` = actual column count (small)
2. `mapping[I]` = index computed against baseSchema column count (large)
3. `mapping[i] - num_input_columns` overflows the `expressions()` array bounds
4. The protobuf repeated field access returns garbage memory
5. Dereferencing that garbage as an Expression pointer causes SIGSEGV

### To Reproduce

Create a plan declares a baseSchema with 10 columns for a table named orders, with `emit.outputMapping = [10]` (meaning: project expression at index `10 - 10 = 0`).

At runtime, the actual orders table has only 4 columns. The extension computes `expressions[10 - 4] = expressions[6]`, which is out of bounds (4 columns, plus only 1 expression), reading garbage memory and crashing.

## Crash Stack Trace
```
Thread 0 Crashed:
0  substrait.duckdb_extension  TransformExpr(Expression const&, RootNameIterator*) + 36
1  substrait.duckdb_extension  TransformProjectOp(Rel const&, ...) + 948
2  substrait.duckdb_extension  TransformJoinOp(Rel const&) + 248
3  substrait.duckdb_extension  TransformJoinOp(Rel const&) + 176
4  substrait.duckdb_extension  TransformJoinOp(Rel const&) + 176
5  substrait.duckdb_extension  TransformProjectOp(Rel const&, ...) + 440
6  substrait.duckdb_extension  TransformAggregateOp(Rel const&) + 120
7  substrait.duckdb_extension  TransformSortOp(Rel const&, ...) + 472
8  substrait.duckdb_extension  TransformProjectOp(Rel const&, ...) + 440
9  substrait.duckdb_extension  TransformRootOp(RelRoot const&) + 92
10 substrait.duckdb_extension  TransformPlan() + 396
11 substrait.duckdb_extension  SubstraitBindReplace(...) + 140
Exception: KERN_INVALID_ADDRESS at 0x697369636572703c (ASCII string data being dereferenced as a pointer).
```

### OS:

osx

latest

### DuckDB Version:

1.4.3

### DuckDB Client:

Python

- [x] I have

- [x] I have

thgschwind:
Check if `mapping[i] - num_input_columns` is a valid index.  If the substrait lists more columns than the actual table has (i.e., the schemas mismatch), this index can be invalid which leading to a segmentation fault.  This OR fixes this issue by adding a bound check and throwing an `InvalidInputException` if it fails.

Fixes: #190

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -692,7 +692,14 @@ SubstraitToDuckDB::TransformProjectOp(const substrait::Rel &sop,
 			if (mapping[i] < num_input_columns) {
 				expressions[i] = make_uniq<PositionalReferenceExpression>(mapping[i] + 1);
 			} else {
-				expressions[i] = TransformExpr(sop.project().expressions(mapping[i] - num_input_columns), &iterator);
+				auto expr_idx = mapping[i] - num_input_columns;
+				if (expr_idx >= (size_t)sop.project().expressions_size()) {
+					throw InvalidInputException(
+					    "Project emit mapping references expression index %d, but only %d expressions exist "
+					    "(possible schema mismatch: plan declares more columns than actual table)",
+					    expr_idx, sop.project().expressions_size());
+				}
+				expressions[i] = TransformExpr(sop.project().expressions(expr_idx), &iterator);
 			}
 		}
 	}
--- a/test/sql/test_schema_mismatch_crash.test
+++ b/test/sql/test_schema_mismatch_crash.test
@@ -1,0 +1,68 @@
+# name: test/sql/test_schema_mismatch_crash.test
+# description: Test that schema mismatch in project emit mapping raises an error instead of crashing (issue #190)
+# group: [sql]
+
+require substrait
+
+statement ok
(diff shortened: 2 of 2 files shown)


thgschwind:
Changed the test in the last commit to use the SQL test framework.

andrew-coleman (approved):

Issue #192, opened 2026-06-02
segfault on deprecated substrait types


thgschwind:
### What happens?

`from_substrait` crashes with a segfault when consuming a binary substrait plan that uses a deprecated type field like  `timestamp` (proto field 14), `timestampTz` (field 29), or `time` (field 17). Ideally, these types should be still supported to support older plans and the extension should not crash with a segfault.

The deprecated type fields (`Timestamp`, `TimestampTZ`, `Time`) were excluded from the generated C++ protobuf code during proto compilation. When a binary plan contains one of these types in a cast expression:

1. The C++ protobuf library stores the field in the `UnknownFieldSet` (since it doesn't recognize the field number)
2. `Type::kind_case()` returns `KIND_NOT_SET` (0)
3. `SubstraitToDuckType()` falls through to the `default:` handler
4. The handler calls `Type::GetDescriptor()->FindFieldByNumber(0)->name()`
5. `FindFieldByNumber(0)` returns `nullptr`
6. Dereferencing `nullptr->name()` causes the segfault at address `0x8`

### To Reproduce

Any substrait plan with a cast to the deprecated `timestamp` type triggers the crash. For example, a plan produced by the Python substrait library:

```python
from google.protobuf.json_format import Parse
from substrait import plan_pb2

plan_json = '{"relations":[{"root":{"input":{"project":{"common":{"emit":{"outputMapping":[1]}},"input":{"read":{"common":{"direct":{}},"baseSchema":{"names":["val"],"struct":{"types":[{"string":{"nullability":"NULLABILITY_NULLABLE"}}],"nullability":"NULLABILITY_REQUIRED"}},"namedTable":{"names":["t"]}}},"expressions":[{"cast":{"type":{"timestamp":{"nullability":"NULLABILITY_NULLABLE"}},"input":{"selection":{"directReference":{"structField":{}},"rootReference":{}}},"failureBehavior":"FAILURE_BEHAVIOR_RETURN_NULL"}}]}},"names":["result"]}}],"version":{"minorNumber":85}}'

plan = Parse(plan_json, plan_pb2.Plan())
# Feed plan.SerializeToString() to DuckDB's from_substrait → SIGSEGV
```

## Crash Report (macOS ARM64)

```
Exception Type:    EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x0000000000000008

Thread 0 Crashed:
[...]

PR #193, opened 2026-06-02
feat: added support for deprecated types Timestamp, Time, and TimestampTZ

thgschwind:
Also in case the type is not recognized the null pointer returned by `substrait::Type::GetDescriptor()->FindFieldByNumber(s_type.kind_case())` is handled without causing a segfault.

Fixes issue #192.


nielspardon:
These deprecated fields where removed from Substrait in [v0.88.0](https://github.com/substrait-io/substrait/releases/tag/v0.88.0) since they were deprecated since a long time and the new types PrecisionTimestamp and PrecisionTimestampTZ have been available since [v0.43.0](https://github.com/substrait-io/substrait/releases/tag/v0.43.0). Instead of adding support back, could you consider migrating to the newer Substrait version?

thgschwind:
That makes sense for newly generated plans, and I agree that producers should use the newer precision timestamp/time types rather than the deprecated fields.

However, the immediate issue here is that feeding such a plan to the extension currently causes DuckDB to crash with a segmentation fault. I think that should not happen regardless of whether the input plan is outdated, unsupported, or malformed. The extension should reject the plan with a clear error instead of dereferencing a null descriptor pointer.

From my perspective, the main point of this PR is therefore to make the extension robust against this input and avoid segfaulting DuckDB. The compatibility handling for the deprecated timestamp/time fields is an additional part, but not the core concern.

nielspardon:
> From my perspective, the main point of this PR is therefore to make the extension robust against this input and avoid segfaulting DuckDB. The compatibility handling for the deprecated timestamp/time fields is an additional part, but not the core concern.

Thanks for the explanation. I agree it shouldn't segfault since at least in theory the way we implemented the removal in Substrait protobuf keeps the field indices reserved and old messages should still get parsed but the content would be ignored by the protobuf parser. That's at least my understanding of how this should work.

EpsilonPrime (approved):

Issue #197, opened 2026-06-22
get_substrait_json crashes with Internal Error when plan has no LogicalProjection


thgschwind:
### What happens?

`get_substrait_json` throws an `InternalException` with message `"Root node has more than 1, or 0 children (0) up to reaching a projection node. Type N"` for certain queries where DuckDB's planner does not produce a `LogicalProjection` operator in the plan tree.

### To Reproduce

```python
import duckdb, os

con = duckdb.connect(':memory:', config={'allow_unsigned_extensions': 'true'})
con.load_extension(os.environ["DUCKDB_SUBSTRAIT_EXTENSION_PATH"])

con.sql("""CREATE TABLE fin_customer_risk_profile (
  customer_id INTEGER NOT NULL,
  risk_rating VARCHAR NOT NULL,
  aml_risk_level VARCHAR NOT NULL
)""")

result = con.execute(
    "CALL get_substrait_json(?)",
    ["SELECT customer_id, aml_risk_level FROM fin_customer_risk_profile WHERE risk_rating = 'HIGH'"]
)
print(result.fetchone()[0])
```

### OS:

OSX

### Substrait-Extension Version:

latest

### DuckDB Version:

1.4.3

### DuckDB Client:

python

### Have you tried this on the latest `master` branch?

- [x] I have

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [x] I have

thgschwind:
Fixed with PR #198

Issue #196, opened 2026-06-22
get_substrait_json produces positional root names (#0, #1) instead of column aliases when ORDER BY is combined with expressions


thgschwind:
### What happens?

When a SQL query includes both **expressions in the SELECT clause** (such as `CAST`, arithmetic) and an **ORDER BY** clause, `get_substrait_json` produces positional names (`#0`, `#1`, ...) in the Substrait plan's `root.names` field instead of the actual column aliases.

This causes consumers of the Substrait plan (including `from_substrait`) to lose column name information in query results.

### To Reproduce

```sql
CREATE TABLE suppliers (supplier_id INTEGER NOT NULL, supplier_name VARCHAR NOT NULL);
CREATE TABLE products (product_id INTEGER NOT NULL, supplier_id INTEGER NOT NULL, price DOUBLE NOT NULL, stock_quantity INTEGER NOT NULL);

-- This produces root.names = ['#0', '#1'] instead of ['supplier_name', 'total_inventory_value']
CALL get_substrait_json('SELECT CAST(s.supplier_name AS VARCHAR) AS supplier_name, CAST(SUM(p.price * p.stock_quantity) AS VARCHAR) AS total_inventory_value FROM suppliers s JOIN products p ON s.supplier_id = p.supplier_id GROUP BY s.supplier_name HAVING SUM(p.price * p.stock_quantity) > 500 ORDER BY SUM(p.price * p.stock_quantity) DESC');
```

### OS:

OSX

### Substrait-Extension Version:

latest

### DuckDB Version:

1.4.3

### DuckDB Client:

duckdb

### Have you tried this on the latest `master` branch?

- [x] I have

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [x] I have

thgschwind:
Investing the bug in more detail, it turns out the `#idx` names are produced by DuckDB in `duckdb/src/planner/binder/query_node/plan_select_node.cpp` in the `Binder::CreatePlan(BoundSelectNode &statement)` function. It converts a bound SELECT statement into a logical plan.

  1. FROM clause → scan (line 21)
  2. WHERE → filter (line 30)
  3. GROUP BY / aggregates (line 33-59)
  4. HAVING (line 61)
  5. Window functions (line 69)
  6. SELECT list → projection (line 111) — this carries the real column aliases
  7. ORDER BY / LIMIT → via VisitQueryNode (line 117)
[...]

PR #198, opened 2026-06-22
fix: column names lost in pass-through projections

thgschwind:
When ORDER BY (or similar operators) sit between the output projection and the alias-defining projection, DuckDB inserts a pass-through projection where all expressions are BoundReferenceExpression nodes with positional names (#0, #1, ...). TransformRootOp was not detecting this case, causing Substrait root names to contain positional placeholders instead of the actual column aliases (issue #196).

Additionally, when the optimizer eliminates LogicalProjection entirely (e.g., when NOT NULL constraints allow pushing column selection into the scan), the projection search loop threw an InternalException instead of gracefully handling the missing node (issue #197).

## Changes

**Use plan names as the primary source for output column names**: instead of walking the operator tree to find a projection and extract aliases (which failed for pass-through projections and missing projections), use `plan_names`, the same authoritative names DuckDB computes a priori for SQL output headers.

**Handle CTAS separately**: for CREATE TABLE AS, the planner sets `plan_names = ["Count"]` (DDL semantics), so column names are taken directly from `CreateInfo.columns` instead.

**Remove tree-walking heuristics**: `IsSetOperation`, the `weird_scenario` flag, the while-loop descending through the tree, and the `InternalException` for unknown tree shapes are all eliminated.

**Tests**: add C API tests for CAST+ORDER BY, arithmetic+ORDER BY, and no-projection scenarios; add a SQL test for the no-projection crash.

### DIFF
--- a/src/include/to_substrait.hpp
+++ b/src/include/to_substrait.hpp
@@ -24,8 +24,9 @@
 namespace duckdb {
 class DuckDBToSubstrait {
 public:
-	explicit DuckDBToSubstrait(ClientContext &context, LogicalOperator &dop, bool strict_p)
-	    : context(context), strict(strict_p) {
+	explicit DuckDBToSubstrait(ClientContext &context, LogicalOperator &dop, bool strict_p,
+	                           vector<string> plan_names_p = {})
+	    : context(context), strict(strict_p), plan_names(std::move(plan_names_p)) {
 		TransformPlan(dop);
 	};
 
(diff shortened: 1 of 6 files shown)


thgschwind:
put in draft for the moment, I think the code can be cleaned up a bit ...

thgschwind:
Updated/Edited the PR message to reflect the changes in the latest commit ([d050214](https://github.com/substrait-io/duckdb-substrait-extension/pull/198/commits/d0502146d936a6a1097eab20900e7e94e334c816)).

EpsilonPrime (approved):

PR #202, opened 2026-07-05
Update to DuckDB v1.5.4

EpsilonPrime:
## Summary

Updates the extension from DuckDB v1.4.3 to v1.5.4:

- Bump the `duckdb` submodule to v1.5.4 and `extension-ci-tools` to the head of its v1.5.4 branch; update the CI workflow pins to match.

The bump compiled cleanly but surfaced five runtime breakages, all fixed here:

- **Use `statistics_extended` when available.** The 1.5 table scan registers `statistics_extended` (which takes a `ColumnIndex`) instead of `statistics`, so `TransformTableScanToSubstrait` crashed calling a null function pointer on every `get_substrait`.
- **Support `COMPARE_DISTINCT_FROM`.** DuckDB 1.5 canonicalizes `NOT (x IS NOT DISTINCT FROM y)` into `IS DISTINCT FROM`. Added `is_distinct_from` mappings to the comparison/filter/join-condition switches and the reverse mapping in `from_substrait`.
- **Handle struct extract pushdown.** 1.5 scans can extract struct fields and emit them directly (`ColumnIndexType::PUSHDOWN_EXTRACT`), which a Substrait `ReadRel` cannot express; `SELECT r.a FROM t` silently degraded to reading `r`. The producer now reads the full columns and wraps the read in a `ProjectRel` with nested field references and an emit mapping; the consumer translates nested references into `struct_extract_at`.
- **Transform all children of set operations.** 1.5 plans n-ary UNION nodes; the producer only serialized the first two children, silently dropping the rest (wrong results on TPC-DS Q33). `SetRel` is natively n-ary, so all children are emitted; the consumer folds multiple inputs left-associatively. Multi-child INTERSECT is rejected since DuckDB's semantics don't match `INTERSECTION_PRIMARY`.
- **Disable the new `CommonSubplan` optimizer** during plan generation; it deduplicates repeated subplans into materialized CTEs (e.g. TPC-H Q11), which cannot be expressed in Substrait.

Test expectation updates:

- TPC-DS Q58 and Q72 now succeed; DuckDB 1.5 resolves the column references those tests expected to be ambiguous.

## Testing

- All 46 sqllogic test cases pass (1260 assertions), including the TPC-H and TPC-DS suites with `PRAGMA enable_verification` round-trip checking.
- All 48 C-unit test cases pass (255 assertions), including the deprecated-type tests from #193.


src/to_substrait.cpp
```diff
@@ -2076,6 +2144,40 @@ substrait::Rel *DuckDBToSubstrait::TransformGet(LogicalOperator &dop) {
...
+			auto selection = new substrait::Expression_FieldReference();
+			selection->set_allocated_root_reference(new substrait::Expression_FieldReference_RootReference());
+			auto segment = selection->mutable_direct_reference()->mutable_struct_field();
+			segment->set_field(static_cast<int32_t>(output_positions[i]));
+			auto current = column;
+			while (current->HasChildren()) {
+				D_ASSERT(current->ChildIndexCount() == 1);
+				auto &child = current->GetChildIndex(0);
+				if (!child.HasPrimaryIndex()) {
+					throw NotImplementedException("Field-name based struct extract pushdown is not supported in "
```
bestbeforetoday: Is there a risk that objects created with `new` above are left allocated and leak memory if this throw occurs?

bestbeforetoday (commented):
To my non-expert eye, this looks generally good, and it does load and function on a very simple scenario for me using DuckDB 1.5.4. Just one query in an in-line comment to set my mind at ease.

src/to_substrait.cpp
EpsilonPrime: Thanks, fixed!

EpsilonPrime: I looked for similar issues and found a lot of potential leaks (some even on the success path).  #203 and #204 address these.

src/from_substrait.cpp
```diff
@@ -1244,13 +1261,18 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformSetOp(const substrait::Rel &sop
...
-
-	return make_shared_ptr<SetOpRelation>(std::move(lhs), std::move(rhs), type);
+	// Fold multiple inputs left-associatively: for UNION_ALL this matches the
+	// n-ary semantics exactly, for MINUS_PRIMARY it matches subtracting every
+	// secondary input from the primary input
+	auto result = TransformOp(inputs[0]);
+	for (int input_idx = 1; input_idx < input_count; input_idx++) {
+		auto rhs = TransformOp(inputs[input_idx], input_idx + 1 == input_count ? names : nullptr);
+		result = make_shared_ptr<SetOpRelation>(std::move(result), std::move(rhs), type);
+	}
```
bestbeforetoday, line 1282: Is this behavior also correct for INTERSECTION_PRIMARY? The Substrait spec says that INTERSECTION_PRIMARY, "Returns all records from the primary input that are present in any secondary input, removing duplicates." Will this folding result in INTERSECTION_PRIMARY being mapped to records that are present in _all_ (instead of _any_)secondary inputs?

 The TransformIntersect function in to_substrait.cpp explicitly rejects operations that do not have exactly two children. Maybe to avoid this issue.

src/to_substrait.cpp
```diff
@@ -2025,42 +2050,85 @@ substrait::Rel *DuckDBToSubstrait::TransformGet(LogicalOperator &dop) {
...
+			if (!dget.projection_ids.empty()) {
+				auto struct_item = select->add_struct_items();
+				if (!column->IsRowIdColumn()) {
+					struct_item->set_field(static_cast<int32_t>(column->GetPrimaryIndex()));
+				}
+			} else if (!column->IsRowIdColumn()) {
 				auto struct_item = select->add_struct_items();
-				struct_item->set_field(static_cast<int32_t>(col_idx));
+				struct_item->set_field(static_cast<int32_t>(column->GetPrimaryIndex()));
 			}
```
bestbeforetoday, line 2134: Is it OK to call add_struct_items in the first branch leaving the default (zero) field for a RowId column? The else-if branch avoids calling add_struct_items for RowId columns. I really don't know; just checking!

src/from_substrait.cpp
```diff
@@ -1244,13 +1261,18 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformSetOp(const substrait::Rel &sop
...
-	auto rhs = TransformOp(inputs[1], names);
-
-	return make_shared_ptr<SetOpRelation>(std::move(lhs), std::move(rhs), type);
+	// Fold multiple inputs left-associatively: for UNION_ALL this matches the
+	// n-ary semantics exactly, for MINUS_PRIMARY it matches subtracting every
+	// secondary input from the primary input
+	auto result = TransformOp(inputs[0]);
+	for (int input_idx = 1; input_idx < input_count; input_idx++) {
+		auto rhs = TransformOp(inputs[input_idx], input_idx + 1 == input_count ? names : nullptr);
+		result = make_shared_ptr<SetOpRelation>(std::move(result), std::move(rhs), type);
```
bestbeforetoday: Does `setop_all` need to be explicitly set to `true` here for UNION_ALL to behave correctly?

bestbeforetoday (commented):
Thank you for the updates. A few more questions in-line.

EpsilonPrime:
_Posted by David's coding agent on his behalf._

Thanks for the careful review — all three of the latest questions were on point. Summary of what changed in response (latest commits on this branch):

**1. `UNION ALL` dropping duplicates — fixed (correctness bug).**
The consumer built `SetOpRelation(left, right, type)` with the default `setop_all=false`, which appends a `DISTINCT` modifier, so every `SET_OP_UNION_ALL` round-tripped as a *distinct* union and silently dropped duplicates. It now passes `setop_all=true` for `UNION_ALL`. Added `test/sql/test_union_all_duplicates.test`, verified it **fails without the fix** (row count 6 vs 2) and passes with it. This was latent because no existing test had surviving duplicate rows in a `UNION ALL` (TPC-DS Q33's outer `GROUP BY … SUM` masked it).

**2. n-ary `INTERSECTION_PRIMARY` folding — fixed (correctness bug).**
Exactly as you suspected. `INTERSECTION_PRIMARY` keeps records present in *any* secondary input, but left-associative folding keeps records present in *every* secondary. Folding is only valid for `UNION_ALL` (associative) and `MINUS_PRIMARY` (subtract each secondary from the primary), so the consumer now rejects `INTERSECTION_PRIMARY` with more than two inputs — matching the `TransformIntersect` guard on the producer side (which is indeed there for this reason).

**3. Mask-projection row-id asymmetry — investigated; kept, now documented.**
I tried unifying the two branches to skip row-id columns in both. That **broke `test_substrait_topn` and `test_limit`** with `Positional reference N out of range`. Root cause: under late materialization, a TopN/Limit scan reads the sort column plus a **row id** and re-fetches the rest by row id, so when the scan has a pushed-down projection the mask must emit one struct item per projected column — including a placeholder (default field `0`) for the row id — to stay positionally aligned with the projection. Dropping it misaligns the downstream references. So the asymmetry is load-bearing; I've added a comment explaining why rather than changing the behavior.

Separately, the original memory-leak question is addressed here (plus a broader exception-safety sweep of `to_substrait.cpp` in the follow-up PRs #203 and #204).

All sqllogic (47 cases) and C unit (48 cases) tests pass on the updated branch.

bestbeforetoday (approved):
Thank you for the updates. Looks good to me.

PR #203, opened 2026-07-07
refactor: make relation-node transforms in to_substrait exception-safe

EpsilonPrime:
## Summary

Follow-up to the memory-safety feedback on #202. Many `Transform*` helpers in `to_substrait.cpp` allocate a raw `substrait::Rel`/`RelRoot` (and their `NamedStruct`/`Type_Struct` schemas) with `new`, then run work that can throw — `TransformOp`/`TransformExpr` recursion, `DuckToSubstraitType`, `RegisterFunction`, or a `NotImplementedException` for an unsupported operator — **before** ownership is transferred via `set_allocated_*`/`return`. On the exception path those objects leaked.

`TransformProjection` additionally leaked its `Rel` on the **success** path: it allocated the `Rel` and then early-`return`ed the child transform for a pass-through projection without freeing it.

This converts the relation-node allocations to the `make_uniq` + `.release()` RAII pattern **already used in this file** by `CreateSetOperation` and `TransformDistinct`, releasing only at the point ownership is handed to a parent message or returned. If an exception propagates, everything is freed automatically; the success path is unchanged. `TransformExpressionGet` also drops a redundant `new`/copy/`delete` of a temporary `Expression` by transforming directly into the repeated field.

Functions converted: `TransformFilter`, `TransformProjection`, `TransformTopN`, `TransformLimit`, `TransformOrderBy`, `TransformComparisonJoin`, `TransformAggregateGroup`, `TransformDummyScan`, `TransformEmptyResult`, `TransformTableScanToSubstrait`, `TransformParquetScanToSubstrait`, `TransformGet`, `TransformExpressionGet`, `TransformCrossProduct`, `TransformUnion`, `TransformExcept`, `TransformIntersect`, `TransformCreateTable`, `SetTableSchema`, `TransformInsertTable`, `TransformDeleteTable`, `TransformCTERef`, `TransformRootOp`.

## Scope

This pass covers the **relation-tree nodes** (`Rel`, `RelRoot`) and the **schema structs** (`NamedStruct`, `Type_Struct`) attached to reads/writes — the structural backbone, and where the leak flagged in #202 lived. Expression-level and scalar `Type`-primitive allocations (e.g. in `DuckToSubstraitType` and the `CreateConjunction` lambdas) follow the same pattern and are a **documented follow-up**, kept out of this PR to bound the diff and the double-free risk of a wholesale ownership rewrite.

## Testing

- No behavior change: all sqllogic (46 cases) and C unit (48 cases) tests pass unchanged.
- Tests exercise every converted function on the success path, confirming no double-free from a mistaken `release()`.

## Base

Based on `EpsilonPrime/update-extension-duckdb` (the DuckDB v1.5.4 PR, #202) since these functions include the v1.5.4 changes. Retarget to `main` once #202 merges.

### DIFF
--- a/src/to_substrait.cpp
+++ b/src/to_substrait.cpp
@@ -968,24 +968,24 @@ substrait::Rel *DuckDBToSubstrait::TransformFilter(LogicalOperator &dop) {
 
 	auto &dfilter = dop.Cast<LogicalFilter>();
 
-	auto res = TransformOp(*dop.children[0]);
+	unique_ptr<substrait::Rel> res(TransformOp(*dop.children[0]));
 
 	if (!dfilter.expressions.empty()) {
-		auto filter = new substrait::Rel();
-		filter->mutable_filter()->set_allocated_input(res);
+		auto filter = make_uniq<substrait::Rel>();
+		filter->mutable_filter()->set_allocated_input(res.release());
 		filter->mutable_filter()->set_allocated_condition(
 		    CreateConjunction(dfilter.expressions, [&](const unique_ptr<Expression> &in) {
 			    auto expr = new substrait::Expression();
 			    TransformExpr(*in, *expr);
 			    return expr;
 		    }));
-		res = filter;
+		res = std::move(filter);
 	}
 
 	if (!dfilter.projection_map.empty()) {
-		auto projection = new substrait::Rel();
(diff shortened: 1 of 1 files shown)


nielspardon (approved):
LGTM

PR #204, opened 2026-07-07
refactor: make expression/type transforms in to_substrait exception-safe

EpsilonPrime:
## Summary

Follow-up to #203, completing the exception-safety pass in `to_substrait.cpp`. #203 converted the relation-tree nodes (`Rel`/`RelRoot`) and their schema structs; this addresses the **remaining** raw protobuf allocations at the expression and scalar-type level. These leaked on the exception path when a nested `TransformExpr` / `DuckToSubstraitType` / `RegisterFunction` call threw before ownership was transferred via `set_allocated_*`/`return`.

## Changes

- **`DuckToSubstraitType`**: build every Substrait type into the stack-local `s_type` via `mutable_*` accessors instead of `new` + `set_allocated_*`. The scalar cases were already leak-free but are converted for uniformity; the important fix is `STRUCT`/`MAP`/`LIST`, whose recursive children could throw and leak the partially-built type.
- **Expression builders returning an owned `Expression*`** (the `Transform*Filter` helpers, `TransformJoinCond`, and both `CreateConjunction` lambdas): now use `make_uniq` + `.release()`.
- **In-place expression builders** (literal decimal, `CASE`, field reference, function argument, extension URN, and the `TransformGet` mask projection): use `mutable_*` accessors and drop their `new` allocations (and the mask projection drops a manual `delete` on the empty path).
- **Dead code / guaranteed leak removed**: `TransformStructExtractFilter` built a field-reference `Expression` that was never used — the function returns the recursively transformed child filter — so it leaked that object on every call. Removed.
- Replaced two `new string()` + `set_allocated_*` with direct setters.

Net effect: **zero raw `new substrait::` allocations remain** in the file (−64 lines).

## Testing

- No behavior change: all sqllogic (46 cases) and C unit (48 cases) tests pass unchanged.
- Since the tests exercise every converted builder on the success path, a mistaken `release()` (double-free) would crash them; none did.

## Base

Stacked on #203 (`EpsilonPrime/to-substrait-exception-safety`). Retarget to `main` once #203 and #202 land.

### DIFF
--- a/src/to_substrait.cpp
+++ b/src/to_substrait.cpp
@@ -81,9 +81,8 @@ string DuckDBToSubstrait::SerializeToJson() const {
 
 void DuckDBToSubstrait::AllocateFunctionArgument(substrait::Expression_ScalarFunction *scalar_fun,
                                                  substrait::Expression *value) {
-	auto function_argument = new substrait::FunctionArgument();
+	auto function_argument = scalar_fun->mutable_arguments()->Add();
 	function_argument->set_allocated_value(value);
-	scalar_fun->mutable_arguments()->AddAllocated(function_argument);
 }
 
 string GetRawValue(hugeint_t value) {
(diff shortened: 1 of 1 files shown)


nielspardon (approved):
LGTM

PR #206, opened 2026-07-08
fix: extension URN bug

### LINKED ISSUE #205: Bug: incomplete extensionUrn for certain functions
### What happens?

Substrait plans generated by the DuckDB substrait extension produce an incomplete
`extensionUrn` value (`"extension:io.substrait:"`). Typically this is the case for comparison functions
(`equal`, `lt`, `gt`, `lte`, `gte`, `not_equal`, `between`, etc.) when operating
on non-boolean types (integers, floats, strings, timestamps, etc.) but can happen for other overloaded functions as well.

## Symptoms

A query like:

```sql
SELECT * FROM t WHERE a = b   -- a, b are INTEGER
```

produces:

```json
"extensionUrns": [
  {
    "extensionUrnAnchor": 1,
    "urn": "extension:io.substrait:"
  }
]
```

## Root Cause

In `src/custom_extensions.cpp`, the function `InsertAllFunctions` is recursive and
receives `file_path` by non-const reference (`string &file_path`). At the base case
(line 55/66/68), it uses `std::move(file_path)` to store the extension path into the
map entry.

For functions registered with `any1` types (e.g., `equal` with `{"any1", "any1"}`),
the recursion expands to all 21x21 = 441 type combinations. The first base case
(typically `bool_bool`, since `"bool"` is first in `GetAllTypes()`) moves from
`file_path`, leaving it empty. All subsequent 440 combinations store an empty
`extension_path`.

When `GetExtensionURN()` is later called for e.g. `equal:i64_i64`:

```cpp
return "extension:io.substrait:" + extension_path;  // extension_path is ""
```

This produces the broken URN `"extension:io.substrait:"`.

### To Reproduce

Run this program:
```python
#!/usr/bin/env python3
"""
Minimal reproducer for the extensionUrn bug.
Generates a substrait plan for an integer comparison and prints the extensionUrns.

With the bug: shows "extension:io.substrait:" (missing functions_comparison).
After fix:    shows "extension:io.substrait:functions_comparison".
"""

import json
import subprocess
from pathlib import Path

duckdb = str(Path(__file__).resolve().parent.parent / "build" / "release" / "duckdb")

sql = """
CREATE TABLE t (a INTEGER, b INTEGER);
INSERT INTO t VALUES (1, 1), (2, 3);
CALL get_substrait_json('SELECT * FROM t WHERE a = b');
"""

out = subprocess.run([duckdb, "-json"], input=sql, capture_output=True, text=True)
plan = json.loads(json.loads(out.stdout)[0]["Json"])

print("extensionUrns:")
for urn in plan.get("extensionUris", plan.get("extensionUrns", [])):
    anchor = urn.get("extensionUriAnchor", urn.get("extensionUrnAnchor"))
    uri = urn.get("uri", urn.get("urn"))
    print(f"  {anchor}: {uri}")
```

### OS:

osx

latest

### DuckDB Version:

1.4.3

### DuckDB Client:

Python

- [x] I have

- [x] I have

thgschwind:
This fixes bug #205 where extension URNs may be incomplete.  This PR replaces `std::move(file_path)` with `file_path` (plain copy) in all three insertion sites in `InsertAllFunctions`.  Without that replacement the first move call removes the value from the original object and hence the extension urn will no longer be available in subsequent invocations.

Fixes: #205, #207


thgschwind:
Looks like there is more to this, one test fails, and I saw some date -> unknown conversion, probably the fix is more complicated than this.

On top, it seems generating `equal:i64_i64` is not substrait spec compliant but should be `equal:any_any` anyway.

thgschwind:
Latest commit fixes the unknown issue that surfaced by the previous commit which is related to the function expansion that caused the urn to become incomplete (due to the `move` function).  Also it fixes since it affects the very same code the use of the concrete function signatures instead of the canonical function names.  See issue #207 for details.

nielspardon:
good catch, I also ran into that issue in #210. since we just merged that PR, could you rebase your fix on latest main?

nielspardon (changes requested):
Thanks for tackling this! The core `std::move(file_path)` → copy fix for #205 is correct, and deriving the compound name from the *declared* YAML types is the right approach for #207.

I verified the `any`-type handling against the [Substrait spec](https://substrait.io/extensions/#any-types) and it's **spec-correct**: `any`/`any1`/`anyN` all collapse to the short name `any` with the numeric suffix dropped (spec's own example: `sum(any1) => sum:any`), so `equal:any_any`, `gte:any_any`, `count:any` now match what substrait-java and the other producers emit. 👍

A couple of things worth addressing before merge, plus some out-of-scope follow-ups.

### Needs a change in this PR

**1. The `is_fallback` guard never fires.** It tests `file_path == "unknown"`, but the token the generator emits is the full URN `extension:io.substrait:unknown` (`custom_extensions_generated.cpp` ~L172–181). So `is_fallback` is always `false` and the check collapses to an unconditional overwrite — identical to the previous behavior. This means the "`unknown` fallback overwrites earlier specific registrations" problem you referenced from #207 (e.g. `min:date` registered under `functions_datetime` getting clobbered to the `unknown` URN) isn't actually fixed. No test covers a datetime combo, which is why CI stays green.

**2. `unknown → any` in `ComputeCanonicalName` fabricates anchors.** `unknown` isn't a Substrait type — it comes from `extensions/unknown.yaml`, which is an example/placeholder extension that substrait-java explicitly excludes from its default collection. Mapping it to `any` yields names like `multiply:any_any` / `sum:any` on the bogus `extension:io.substrait:unknown` URN, but `add`/`multiply`/`sum`/… are canonically declared over concrete types in the real YAMLs, so those wildcard names correspond to no real signature.

Both have the same clean fix: **exclude `unknown.yaml` from generation** (the generator currently walks the whole extensions dir and ingests it). That lets you delete the `is_fallback` block *and* the `unknown →` branch entirely, makes uncovered type combos fall back to `native` instead of a fake URN, and genuinely resolves #207's sub-issue. A `min(date)` / `date + interval` regression test would lock it in. If you'd prefer to keep that sub-issue for a separate PR, I'd at least drop the non-functional `is_fallback` code rather than merge it as-is.

Minor: `auto canonical = function.GetCanonicalName()` copies (getter returns `const string&`) — `const auto&` avoids it; and `ComputeCanonicalName` only handles `any1`/`any2` while the spec allows `any0`–`any9` (harmless today since only those two are declared).

### Follow-ups (out of scope here)

- **Short-name abbreviation:** the spec uses abbreviated type short names (`decimal→dec`, `string→str`, `binary→vbin`, `precision_timestamp→pts`, …), so `multiply:decimal_decimal` / `gte:string_string` should really be `multiply:dec_dec` / `gte:str_str`. Pre-existing and broader — probably its own issue.
- **Generator hygiene:** skip example/placeholder extension YAMLs in general. `functions_geometry` and `functions_aggregate_decimal_output` are real and should stay.

Happy to help with any of this.

nielspardon:
Status update now that #225 (short names) and #239 (skip example/placeholder extension YAMLs, i.e. `unknown.yaml`) have both landed on `main`:

**Still needed from this PR:**
1. The core `std::move(file_path)` → copy fix for the empty-URN bug (#205). Nothing else covers it — `equal(i32,i32)` still serializes with an empty URN after #239. The regression test added here in `test_substrait_c_api.cpp` remains valuable.
2. `ComputeCanonicalName` — emitting the declared YAML signature (`equal:any_any`, `gte:any_any`) instead of the concrete expanded types (#207's main issue). #225 only *shortened* the concrete names; it didn't switch to the spec signature. This will need a rebase onto #225 so the canonical name also runs through the short-name mapping (e.g. `avg:dec`, not `avg:decimal`).

**Now redundant (handled by #239):**
- DECIMAL casing: #239 normalizes `DECIMAL`→`decimal` at load time (covers sum/avg/min/max/bitwise_*/sqrt/factorial/power/sum0), so the generator-side `.lower()` + regeneration isn't needed.
- The `is_fallback == "unknown"` anti-clobber guard: #239 drops `unknown.yaml` entirely, removing the clobbering source (#207 related issue #1), so there's nothing left to guard against.

Heads-up: #239 and this PR both rewrite `InsertAllFunctions` and `test/c/test_projection.cpp`, so a rebase will be needed to bring this up to date with `main`.

🤖 Generated with AI

andrew-coleman:
Superseded by https://github.com/substrait-io/duckdb-substrait-extension/pull/251

Thanks @thgschwind :)

PR #225, opened 2026-07-16
fix: emit Substrait short names in compound function signatures

### LINKED ISSUE #213: Compound function signatures use full type names instead of Substrait short names (e.g. multiply:decimal_decimal should be multiply:dec_dec)
Follow-up from #206 / #207.

#206 correctly fixed the `any`-type case (`equal:i32_i32` → `equal:any_any`) by deriving the compound name from the declared YAML types. However, **concrete and parameterized argument types are still emitted with their full names instead of the Substrait short names.**

### Problem

Per the [Substrait "Type Short Names" table](https://substrait.io/extensions/), compound function signatures are built exclusively from abbreviated short names. The extension currently emits full type names for everything except the `any` family and `boolean`.

| Emitted now | Spec-compliant |
|---|---|
| `multiply:decimal_decimal` | `multiply:dec_dec` |
| `gte:string_string` | `gte:str_str` |
| `...:binary` | `...:vbin` |
| `...:varchar` | `...:vchar` |
| `...:precision_timestamp` | `...:pts` |

Short-name mappings that currently pass through verbatim (and are therefore wrong):

- `string` → `str`
- `binary` → `vbin`
- `decimal<P,S>` → `dec`
- `varchar<N>` → `vchar`
- `fixedchar<N>` → `fchar`
- `fixedbinary<N>` → `fbin`
- `interval_year` → `iyear`, `interval_day` → `iday`, `interval_compound` → `icompound`
- `precision_time<P>` → `pt`, `precision_timestamp<P>` → `pts`, `precision_timestamp_tz<P>` → `ptstz`

Types that already map to themselves are fine: `i8`/`i16`/`i32`/`i64`, `fp32`/`fp64`, `date`, `uuid`, `struct`/`list`/`map`/`func`, and `any`.

(Note: `timestamp`/`timestamp_tz`/`time` no longer appear in the current spec table — they've been superseded by the `precision_*` variants.)

### Root cause

`ComputeCanonicalName` (and the older `SubstraitCustomFunction::GetName`) join the declared argument tokens with `_`, special-casing only `any*`→`any` and `boolean`→`bool`. They need the full Type Short Names lookup table.

### Impact

Signatures don't match the spec or other producers (substrait-java/go/python). Round-trips within the DuckDB extension itself are unaffected (the reader's `RemoveExtension` strips everything after `:`), but external Substrait consumers and spec-compliance are.

### Scope note

This affects the large majority of declared argument tokens (`string` ×107, `varchar` ×74, `interval_year` ×23, `decimal` ×20, …), so it's a broad change touching many test expectations — hence tracked separately from #206.

nielspardon:
Fixes #213.

## Problem

Compound function signatures were built by joining the declared argument tokens verbatim, so the extension emitted full type names for everything except `boolean` and the `any` family — e.g. `multiply:decimal_decimal`, `gte:string_string` — instead of the abbreviated [Substrait Type Short Names](https://substrait.io/extensions/). This diverges from the spec and from other producers (substrait-java/go/python).

## Fix

`SubstraitCustomFunction::GetName()` now maps each argument token through the Substrait Type Short Names table before joining:

| Before | After |
|---|---|
| `multiply:decimal_decimal` | `multiply:dec_dec` |
| `gte:string_string` | `gte:str_str` |
| `equal:binary_binary` | `equal:vbin_vbin` |
| `equal:precision_timestamp_tz_...` | `equal:ptstz_ptstz` |
| `and:bool?` | `and:bool` |

Full mapping covers `string→str`, `binary→vbin`, `decimal→dec`, `varchar→vchar`, `fixed_char→fchar`, `fixed_binary→fbin`, `interval_year→iyear`, `interval_day→iday`, `interval_compound→icompound`, `precision_time→pt`, `precision_timestamp→pts`, `precision_timestamp_tz→ptstz`. Types whose short name equals their name (`i8`..`i64`, `fp32`/`fp64`, `bool`, `date`, `uuid`, `struct`/`list`/`map`, `func`, `any`) pass through unchanged.

The trailing `?` nullability marker on variadic arguments is now stripped, since Substrait compound names encode only the short type name (so the variadic `and` emits `and:bool`, not `and:bool?`).

This only affects the producer-side signature string. Round-trips within the extension are unaffected (the reader strips everything after `:`); the change is for spec compliance and interoperability with external Substrait consumers.

## Tests

Updated affected expectations in `test/sql/test_custom_function.test`, `test/sql/test_substrait_enable_optimizer.test`, and `test/c/test_projection.cpp`. Full SQL suite (48 cases / 1268 assertions) and C++ suite (48 cases / 255 assertions) pass.

🤖 Generated with AI


src/custom_extensions.cpp
```diff
@@ -101,13 +101,49 @@ void SubstraitCustomFunctions::InsertCustomFunction(string name_p, vector<string
...
+// whose short name is identical to their name (i8/i16/i32/i64, fp32/fp64, bool,
+// date, uuid, struct/list/map, func, any) are absent from the table and pass
+// through unchanged.
+static string TypeShortName(const string &type) {
+	static const std::unordered_map<string, string> SHORT_NAMES = {
+	    {"string", "str"},
+	    {"binary", "vbin"},
+	    {"decimal", "dec"},
+	    {"varchar", "vchar"},
+	    {"fixed_char", "fchar"},
```
EpsilonPrime, line 125: There's a fixedchar in the functions_string.yaml file so it's likely this won't trigger.   Add fixedchar as well?

EpsilonPrime (approved):

src/custom_extensions.cpp
nielspardon: Good catch. Rather than special-casing it in the short-name table, I hardened the extension→proto type normalization in `InsertAllFunctions` (added `fixedchar` → `fixed_char` alongside the existing `boolean` → `bool`, plus `fixedbinary` → `fixed_binary`), so the stored overload tokens now match the protobuf `Type.kind` names that `TransformTypes()` produces at lookup. The `fixed_char` → `fchar` mapping then resolves correctly if a `FixedChar` argument ever reaches the producer.

It stays dormant today: DuckDB has no fixed-length char type — `CHAR(N)` is `VARCHAR` → substrait `string` — so the producer never emits `FixedChar` (verified: `upper()` on a `CHAR(5)` column emits `upper:str`).

Filed #227 for the broader version of this: the `any`-expansion list in `GetAllTypes()` is out of sync with the kinds the producer can actually emit.

PR #239, opened 2026-07-21
fix: only ingest functions_*.yaml extension definitions (#214)

### LINKED ISSUE #214: Code generator ingests example-only extension YAMLs (unknown.yaml), yielding bogus URNs and clobbering real registrations
Follow-up from #206 / #207.

### Problem

`scripts/generate_custom_functions.py` walks the entire extensions directory and ingests every YAML it finds. That includes [`extensions/unknown.yaml`](https://github.com/substrait-io/substrait/blob/main/extensions/unknown.yaml), which is an **example/placeholder** extension (`urn: extension:io.substrait:unknown`) that declares a custom type literally named `unknown` and defines `add`/`subtract`/`multiply`/`divide`/`modulus`/`sum`/`avg`/`min`/`max`/`count` over it.

Other producers deliberately exclude it — substrait-java loads only an explicit `functions_*.yaml` allowlist (plus `extension_types.yaml`).

### Consequences

1. **Clobbering:** `InsertCustomFunction` treats the custom `unknown` type as a wildcard and expands it via `GetAllTypes()`, producing catch-all entries that can overwrite earlier, correct registrations (e.g. `min:date` registered under `functions_datetime` gets overwritten with the `unknown` URN).
2. **Bogus URNs / fabricated signatures:** these functions carry the placeholder URN `extension:io.substrait:unknown`, and (after #206) `unknown` is mapped to `any`, producing names like `multiply:any_any` / `sum:any` that correspond to no real declared signature.

### Suggested fix

Make extension selection explicit — skip example/placeholder extensions (an allowlist of `functions_*.yaml` + the needed type files, or a deny-list containing `unknown.yaml`). Type combinations not covered by a real extension then correctly fall back to `native` instead of referencing a fake URN.

Keep `functions_geometry` and `functions_aggregate_decimal_output` — those are legitimate extensions the DuckDB extension supports.

### Relationship to #206

#206 may address the immediate `unknown.yaml` instance directly (suggested in review). This issue tracks the general hygiene: ensure the generator never silently ingests placeholder / non-function extension YAMLs.

nielspardon:
Fixes #214.

## Problem

`scripts/generate_custom_functions.py` walked the entire Substrait extensions directory and ingested **every** YAML it found — including [`extensions/unknown.yaml`](https://github.com/substrait-io/substrait/blob/main/extensions/unknown.yaml), an example/placeholder extension (`urn: extension:io.substrait:unknown`) that declares a custom type literally named `unknown` and defines `add`/`subtract`/`multiply`/`divide`/`modulus`/`sum`/`avg`/`min`/`max`/`count` over it.

Consequences:

1. **Clobbering:** `InsertCustomFunction` treats the custom `unknown` type as a wildcard and expands it via `GetAllTypes()`, producing catch-all entries that can overwrite earlier, correct registrations (e.g. `min:date` registered under `functions_datetime` gets overwritten with the `unknown` URN).
2. **Bogus URNs / fabricated signatures:** these functions carried the placeholder URN `extension:io.substrait:unknown` and produced names like `multiply:any_any` / `sum:any` that correspond to no real declared signature.

## Fix

Make extension selection explicit: allowlist by the `functions_*.yaml` naming convention, mirroring substrait-java. This skips:

- example/placeholder extensions (`unknown.yaml`), and
- type-only definitions (`extension_types.yaml`, `type_variations.yaml`) that declare no functions.

Type combinations not covered by a real extension now correctly fall back to `native` instead of referencing a fake URN. `functions_geometry` and `functions_aggregate_decimal_output` remain ingested, as requested in the issue. Skipped files are logged during generation for transparency.

`src/custom_extensions_generated.cpp` was regenerated from `substrait-packaging` tag `cpp/substrait-extensions/v0.94.0` (already pinned in the script). The only change is the removal of the 10 bogus `extension:io.substrait:unknown` registrations.

## Follow-up cleanup

With the generator no longer emitting any `unknown` argument type (and `TransformTypes()` never producing `"unknown"` — the placeholder is a user-defined type at runtime), the `t == "unknown"` expansion branch in `InsertCustomFunction` was unreachable. It and the stale comment reference are removed.

## Latent bug surfaced by dropping `unknown.yaml`

Removing `unknown.yaml` exposed a bug it had been masking. The aggregate `sum`/`avg`/`min`/`max` in `functions_arithmetic_decimal` spell their argument `DECIMAL` (uppercase), while `TransformTypes()` only ever yields the lowercase proto kind name `decimal` at lookup — so those overloads never resolved on their own. The `unknown` catch-all (expanded over `GetAllTypes()`, which includes lowercase `decimal`) was silently standing in for them, which is why e.g. TPC-H Q1 (`sum`/`avg` over decimal) passed before.

This PR normalizes the declared `DECIMAL` → `decimal` alongside the existing spelling normalizations (`boolean`→`bool`, `fixedchar`→`fixed_char`, …), so `sum`/`avg`/`min`/`max` over decimal now resolve to the real `functions_arithmetic_decimal` URN. This closes the `avg:dec` resolution gap that was documented as a known follow-up in `test/c/test_projection.cpp` (its expected JSON and note are updated accordingly).

## Testing

- `build/release/test/unittest 'test/sql/*'` → all pass (49 cases, 1293 assertions; 1 httpfs skip).
- C API tests (`test_substrait_exe`) → all pass (48 cases, 255 assertions; 1 skip).
- The previously-failing `test/sql/test_substrait_tpch.test` (strict-mode `sum`/`avg` over decimal) now passes.

🤖 Generated with AI

### DIFF
--- a/scripts/generate_custom_functions.py
+++ b/scripts/generate_custom_functions.py
@@ -16,6 +16,26 @@ SUBSTRAIT_PACKAGING_REPO = "https://github.com/substrait-io/substrait-packaging.
 SUBSTRAIT_EXTENSIONS_TAG = "cpp/substrait-extensions/v0.94.0"
 SUBSTRAIT_EXTENSIONS_SUBDIR = os.path.join("cpp", "substrait-extensions", "extensions")
 
+# Only ingest extension YAMLs that declare real Substrait functions. Every
+# legitimate function extension follows the `functions_*.yaml` naming
+# convention, so we allowlist by that prefix rather than walking every YAML in
+# the directory. This deliberately skips:
+#   * example/placeholder extensions -- notably `unknown.yaml`, which declares a
+#     custom `unknown` type and defines add/subtract/.../count over it under the
+#     placeholder URN `extension:io.substrait:unknown`. Ingesting it produced
+#     fabricated signatures and a bogus URN that could clobber real
+#     registrations (see issue #214).
+#   * type-only definitions (`extension_types.yaml`, `type_variations.yaml`),
+#     which declare no functions.
+# This mirrors substrait-java, which loads an explicit `functions_*.yaml`
+# allowlist. Type combinations not covered by a real extension then correctly
+# fall back to `native` instead of referencing a fake URN.
+FUNCTION_EXTENSION_PREFIX = "functions_"
+
+
+def is_function_extension(file_name):
+	return file_name.startswith(FUNCTION_EXTENSION_PREFIX) and file_name.endswith(".yaml")
+
 
 def parse_function_data(functions,yaml_data,function_type):
 	for function_data in yaml_data.get(function_type, []):
@@ -51,6 +71,9 @@ def get_custom_functions(custom_extension_folder):
 	type_set = set()
 	custom_function_paths = next(walk(custom_extension_folder), (None, None, []))[2]
 	for custom_function_path in custom_function_paths:
+		if not is_function_extension(custom_function_path):
+			print(f"Skipping non-function extension YAML: {custom_function_path}")
+			continue
 		# Each extension YAML declares its own Substrait URN
 		# (extension:<owner>:<id>); use it directly rather than reconstructing it
 		# from the file name.
--- a/src/custom_extensions.cpp
+++ b/src/custom_extensions.cpp
@@ -21,7 +21,7 @@ string TransformTypes(const substrait::Type &type) {
 	return string(field->name());
 }
 
-// Concrete types over which an `any`/`any1`/`unknown` argument is expanded when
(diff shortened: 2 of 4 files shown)


EpsilonPrime (approved):
I'd argue if folks aren't supposed to read it then we shouldn't put it in core.  Ok to fix this temporarily here however.

nielspardon:
> I'd argue if folks aren't supposed to read it then we shouldn't put it in core. Ok to fix this temporarily here however.

Agreed, I was also thinking about proposing a follow-up to the spec to move those into an examples or test folder.

nielspardon:
spec follow-up is here (with some open questions): https://github.com/substrait-io/substrait/pull/1136

PR #240, opened 2026-07-21
feat: implement support for missing Substrait literal types

andrew-coleman:
Implement support for the following unimplemented literal types in Substrait to DuckDB conversion:
- i16 (signed 16-bit integer)
- binary and fixed_binary
- precision_timestamp and precision_timestamp_tz
- uuid
- fixed_char
- struct, map, list, empty_list, empty_map
- interval_compound

Update test_literals.test to mark these types as supported with expected outputs.
Add extra tests (non-empty lists, maps, structs) for full coverage. User-defined literals are currently unsupported.

Assisted by AI.

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -104,6 +104,31 @@ SubstraitToDuckDB::SubstraitToDuckDB(shared_ptr<ClientContext> &context_p, const
 	}
 }
 
+static const int64_t POWERS_OF_10[] = {
+	1LL,
+	10LL,
+	100LL,
+	1000LL,
+	10000LL,
+	100000LL,
+	1000000LL,
+	10000000LL,
+	100000000LL,
+	1000000000LL
+};
+
+static int64_t ScaleToMicros(int64_t value, int32_t precision) {
+	if (precision < 0 || precision > 9) {
+		throw InvalidInputException("Precision must be between 0 and 9, got %d", precision);
+	}
+	if (precision > 6) {
+		return value / POWERS_OF_10[precision - 6];
+	} else if (precision < 6) {
+		return value * POWERS_OF_10[6 - precision];
+	}
+	return value;
+}
+
 Value TransformLiteralToValue(const substrait::Expression_Literal &literal) {
 	if (literal.has_null()) {
 		return Value(LogicalType::SQLNULL);
@@ -148,6 +173,8 @@ Value TransformLiteralToValue(const substrait::Expression_Literal &literal) {
 	}
 	case substrait::Expression_Literal::LiteralTypeCase::kI8:
 		return Value::TINYINT(static_cast<int8_t>(literal.i8()));
+	case substrait::Expression_Literal::LiteralTypeCase::kI16:
+		return Value::SMALLINT(static_cast<int16_t>(literal.i16()));
 	case substrait::Expression_Literal::LiteralTypeCase::kI32:
 		return Value::INTEGER(literal.i32());
 	case substrait::Expression_Literal::LiteralTypeCase::kI64:
@@ -162,7 +189,7 @@ Value TransformLiteralToValue(const substrait::Expression_Literal &literal) {
 	}
 	case substrait::Expression_Literal::LiteralTypeCase::kIntervalYearToMonth: {
 		interval_t interval {};
-		interval.months = literal.interval_year_to_month().months();
+		interval.months = (int32_t)literal.interval_year_to_month().years() * 12 + literal.interval_year_to_month().months();
 		interval.days = 0;
 		interval.micros = 0;
 		return Value::INTERVAL(interval);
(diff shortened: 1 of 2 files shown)


nielspardon:
## Code Review

### Overview
Fills in the `default → NotImplementedException` gap in `TransformLiteralToValue` for a large set of Substrait literal types: `i16`, `binary`/`fixed_binary`, `precision_timestamp`(`_tz`), `uuid`, `fixed_char`, `struct`/`map`/`list`/`empty_list`/`empty_map`, and `interval_compound`. It also fixes two pre-existing bugs (year-to-month interval ignoring `years`; day-to-second precision scaling), replaces floating-point `pow()` with an integer `POWERS_OF_10` lookup, and flips the corresponding `test_literals.test` cases from `statement error` to `query I` with expected output.

Overall this is a solid, well-tested improvement. The bug fixes and the `pow → integer table` change are clear wins. A few correctness and consistency issues below.

### Correctness

- **UUID returns a `BLOB`, not a `UUID` — inconsistent with the type mapping.** `kUuid` produces `Value::BLOB(...)`, but `SubstraitToDuckType` maps `kUuid → LogicalTypeId::UUID`. These tests pass only because a bare projected literal infers its type from the value (BLOB). In a `virtualTable`/read context where the schema declares a `UUID` column, a BLOB literal would need a BLOB→UUID cast, which DuckDB does not provide implicitly — so real round-trips through a typed schema would fail. Consider converting the 16 big-endian bytes to `Value::UUID` (hugeint), or at minimum document the BLOB limitation and note the type mismatch. There's also **no positive UUID test** — only the 16-byte-length error path is covered.

- **List/map/struct element types are inferred from the *first* element.** `element_type = list_values[0].type()` then `Value::LIST(element_type, ...)` casts every element to that type. If the first element is a NULL literal, `element_type` becomes `SQLNULL` and casting subsequent non-null values (e.g. `[null, 1, 2]`) will throw. Substrait list values carry their own per-element type but no list-level element type here, so picking the first non-null element's type (or falling back gracefully) would be more robust. Homogeneous lists are fine; the mixed-null case is the risk.

- **`empty_list` vs empty `list{}` produce different element types.** `kEmptyList` uses `LogicalType::SQLNULL`, but an empty `kList` (no values) falls through to the `VARCHAR` default. Both render as `[]` so the tests pass, but the resulting LIST types differ. Worth making consistent.

- **Empty struct → `NULL`.** Returning `SQLNULL` for a zero-field struct is a defensible workaround (DuckDB can't represent a 0-field struct), but semantically an empty struct literal isn't null. Fine to keep given the constraint — a comment explaining *why* (DuckDB limitation) would help the next reader.

- **`precision_timestamp` literal is always micros regardless of precision.** The literal scales to microseconds and returns plain `TIMESTAMP`, while `SubstraitToDuckType` maps precision 0/3/9 to `TIMESTAMP_SEC`/`_MS`/`_NS`. Standalone this is fine; combined with a typed schema it's a latent mismatch. Low priority.

### Code quality / style
- `POWERS_OF_10` + `ScaleToMicros` with an explicit 0–9 range check is a clean replacement for `(int64_t)pow(10, ...)` and avoids FP rounding. Good. Max index used is 6, well within bounds.
- **Duplication:** the day-to-second→micros computation and the `years*12 + months` logic each appear twice (`kIntervalDayToSecond` + `kIntervalCompound`, and `kIntervalYearToMonth` + `kIntervalCompound`). Consider extracting small helpers (e.g. `DayToSecondMicros(...)`).
- `(int32_t)literal.interval_year_to_month().years() * 12` — the cast is redundant (`years()` is already `int32`) but harmless.
- Struct field naming `"f" + std::to_string(i)` matches the `SubstraitToDuckType` convention — nice consistency.
- Minor: most `throw` messages use printf-style (`"...%d"`), but the UUID/struct messages use `std::to_string` concatenation (the existing decimal check does too, so not out of place).

### Performance
Negligible. The lookup table is strictly better than repeated `pow()`. Recursive descent for nested literals is fine for realistic depths.

### Security
- Base64 test inputs were corrected (`"AAEC"` → `\x00\x01\x02`); the old `"XDAwXDAxXDAy"` decoded to 9 bytes, which now usefully drives the UUID length-check test.
- The 16-byte UUID guard mirrors the existing decimal guard — good defensive input validation. No new injection/overflow surface.

### Suggestions (priority order)
1. Resolve the UUID BLOB-vs-UUID inconsistency (convert to `Value::UUID`, or document + add a positive test).
2. Guard the first-element type inference against a leading NULL in lists/maps.
3. Make empty `list{}` and `empty_list` yield the same element type.
4. Factor out the duplicated interval-conversion logic.
5. Optionally add tests for out-of-range precision (the `ScaleToMicros` throw) and a valid 16-byte UUID.

None of these are blocking except the UUID type mismatch, which I'd want acknowledged as a known limitation before merge.

🤖 Generated with AI

nielspardon:
## Re-review (after commit `2e5830a1`)

Thanks for the quick turnaround — the refactor addresses the earlier feedback. Delta review below.

### Resolution of previous feedback

| # | Previous point | Status |
|---|---|---|
| 1 | UUID returns BLOB, not UUID | **Documented + test added.** Comment now explains the BLOB→UUID limitation; a positive 16-byte UUID test was added. Took the "document + test" branch rather than a true UUID conversion — acceptable. |
| 2 | Type inferred from first element (leading-NULL bug) | **Fixed.** Both list and map now skip `SQLNULL` elements to find a concrete type, defaulting to VARCHAR if all-null. |
| 3 | `empty_list` vs empty `list{}` type mismatch | **Fixed.** `kEmptyList` now uses `VARCHAR`, consistent with the empty `kList` path. |
| 4 | Duplicated interval logic | **Fixed.** Extracted `DayToSecondMicros(...)` and `YearToMonthMonths(...)` helpers, used by both the standalone and compound cases. |
| 5 | Empty struct → NULL (needs a why) | **Addressed.** Comment now explains DuckDB can't represent a 0-field struct. |

The helpers read well and the change is clean.

### Remaining / new observations

- **The leading-NULL fix is itself untested.** The inference loop that skips `SQLNULL` is the actual bug fix from point #2, but no test exercises a list/map with a leading null (e.g. `[null, 1, 2]`). Without a case that hits the skip path, a regression here wouldn't be caught. Worth one small test. Note this only fixes *type inference* — `Value::LIST`/`Value::MAP` still cast every element to the inferred type, so a genuinely heterogeneous list would still throw (fine, since Substrait lists are homogeneous; just confirming scope).

- **`ScaleToMicros` out-of-range precision throw is still uncovered** (flagged as optional). Precisions >9 / <0 hit `InvalidInputException` but no test drives it.

- **`precision_timestamp` value-vs-type mismatch** (previously low priority) is unchanged: the literal always yields micros `TIMESTAMP`, while `SubstraitToDuckType` maps precision 0/3/9 to `TIMESTAMP_SEC`/`_MS`/`_NS`. Still fine for standalone literals; still a latent mismatch under a typed schema. Not blocking.

- **New helpers look correct.** `DayToSecondMicros` casts `seconds` to `int64_t` before multiplying (no overflow), `subseconds` is `int64_t`, and precision validation is centralized in `ScaleToMicros`. `YearToMonthMonths` is a plain `years*12 + months` (theoretical int32 overflow only at unrealistic magnitudes).

- **New UUID test is self-consistent** — `"AAAA…AA=="` decodes to 16 zero bytes → rendered as sixteen `\x00`.

### Verdict
All five review points are resolved or consciously deferred. The only thing I'd still nudge for is a test covering the leading-NULL inference path (since that was the actual correctness fix). The UUID-as-BLOB semantic remains a documented known limitation rather than a true fix — acceptable if maintainers are comfortable with the trade-off. Otherwise this looks ready to merge.

🤖 Generated with AI

PR #245, opened 2026-07-22
build: fix Substrait 0.98.0 breaking API changes

### LINKED ISSUE #234: Update to Substrait v0.98.0
# Release Notes

### ⚠ BREAKING CHANGES

* **protos:** The deprecated `left_keys` and `right_keys` fields have
been removed from `HashJoinRel` and `MergeJoinRel`. Producers must use
the `keys` field instead.
* **protos:** The deprecated `offset` and `count` fields have been
removed from `FetchRel`. Producers must use `offset_expr` and
`count_expr` instead.
* **protos:** remove deprecated VirtualTable.values field (#1131)
* **protos:** remove deprecated IntervalDayToSecond.microseconds field (#1116)

### Features

* **protos:** remove deprecated IntervalDayToSecond.microseconds field ([#1116](https://github.com/substrait-io/substrait/issues/1116)) ([a4ad9f4](https://github.com/substrait-io/substrait/commit/a4ad9f4208e827be845ecf3e307f58277e2cfc0f))
* **protos:** remove deprecated left_keys/right_keys from join rels ([#1129](https://github.com/substrait-io/substrait/issues/1129)) ([08b2736](https://github.com/substrait-io/substrait/commit/08b27364728822bf00bf13a82998f09f1c4b96d7))
* **protos:** remove deprecated offset/count from FetchRel ([#1130](https://github.com/substrait-io/substrait/issues/1130)) ([b7e8d02](https://github.com/substrait-io/substrait/commit/b7e8d027e01e21656ab498d23db4335098f50745))
* **protos:** remove deprecated VirtualTable.values field ([#1131](https://github.com/substrait-io/substrait/issues/1131)) ([7b0bd66](https://github.com/substrait-io/substrait/commit/7b0bd664b5e09bf6a230742b17511716040a9d21))

### Bug Fixes

* **tests:** correct decimal sum and nullif results ([#1048](https://github.com/substrait-io/substrait/issues/1048)) ([a750c25](https://github.com/substrait-io/substrait/commit/a750c25c3bd3c9e82fb10f1c1bbf3d0a273a0f46))

See https://github.com/substrait-io/substrait/releases/tag/v0.98.0

andrew-coleman:
Replace deprecated FetchRel setter methods with Expression-based API:
- set_offset/set_count → offset_expr/count_expr
- Extract integer values from expressions with has_literal() checks

Replace deprecated VirtualTable.values() with expressions():
- Use add_expressions() instead of add_values()
- Extract literals from Expression.fields() instead of direct access

Update test JSON format to use new expression-based API for FetchRel, with proper field naming for JSON deserialization (offsetExpr/countExpr).

Resolves #234

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -199,6 +199,17 @@ unique_ptr<ParsedExpression> SubstraitToDuckDB::TransformLiteralExpr(const subst
 	return make_uniq<ConstantExpression>(TransformLiteralToValue(sexpr.literal()));
 }
 
+static idx_t ExtractLiteralInteger(const substrait::Expression &expr, const char *context_desc) {
+	if (!expr.has_literal()) {
+		throw NotImplementedException("Non-literal expressions in %s are not supported", context_desc);
+	}
+	auto val = TransformLiteralToValue(expr.literal());
+	if (val.IsNull()) {
+		throw NotImplementedException("NULL expressions in %s are not supported", context_desc);
+	}
+	return val.GetValue<int64_t>();
+}
+
 unique_ptr<ParsedExpression> SubstraitToDuckDB::TransformSelectionExpr(const substrait::Expression &sexpr) {
 	if (!sexpr.selection().has_direct_reference() || !sexpr.selection().direct_reference().has_struct_field()) {
 		throw SyntaxException("Can only have direct struct references in selections");
@@ -630,8 +641,17 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformCrossProductOp(const substrait:
 shared_ptr<Relation> SubstraitToDuckDB::TransformFetchOp(const substrait::Rel &sop,
                                                          const google::protobuf::RepeatedPtrField<std::string> *names) {
 	auto &slimit = sop.fetch();
-	idx_t limit = slimit.count() == -1 ? NumericLimits<idx_t>::Maximum() : slimit.count();
-	idx_t offset = slimit.offset();
+
+	idx_t limit = NumericLimits<idx_t>::Maximum();
+	if (slimit.has_count_expr()) {
+		limit = ExtractLiteralInteger(slimit.count_expr(), "LIMIT count");
+	}
+
+	idx_t offset = 0;
+	if (slimit.has_offset_expr()) {
+		offset = ExtractLiteralInteger(slimit.offset_expr(), "LIMIT offset");
+	}
+
 	return make_shared_ptr<LimitRelation>(TransformOp(slimit.input(), names), limit, offset);
 }
 
@@ -711,7 +731,7 @@ SubstraitToDuckDB::TransformProjectOp(const substrait::Rel &sop,
 		auto &sget = sop.project().input().read();
 		if (sget.has_virtual_table()) {
 			auto virtual_table = sget.virtual_table();
-			if ((virtual_table.values().empty() && virtual_table.expressions().empty()) ||
+			if ((virtual_table.expressions().empty()) ||
 			    (virtual_table.expressions().size() > 0 && virtual_table.expressions(0).fields().empty())) {
 				hasZeroColumnVirtualTable = true;
 				input_rel = GetValueRelationWithSingleBoolColumn();
@@ -927,14 +947,18 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformReadOp(const substrait::Rel &so
 		scan = rel->Alias(name);
 	} else if (sget.has_virtual_table()) {
 		// We need to handle a virtual table as a LogicalExpressionGet
-		if (!sget.virtual_table().values().empty()) {
-			auto literal_values = sget.virtual_table().values();
+		if (!sget.virtual_table().expressions().empty()) {
+			auto literal_values = sget.virtual_table().expressions();
 			vector<vector<Value>> expression_rows;
 			for (auto &row : literal_values) {
 				auto values = row.fields();
 				vector<Value> expression_row;
 				for (const auto &value : values) {
-					expression_row.emplace_back(TransformLiteralToValue(value));
+					if (value.has_literal()) {
+						expression_row.emplace_back(TransformLiteralToValue(value.literal()));
+					} else {
+						throw NotImplementedException("Non-literal expressions in virtual table values are not supported");
+					}
 				}
 				expression_rows.emplace_back(expression_row);
 			}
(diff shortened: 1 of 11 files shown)


src/from_substrait.cpp
```diff
@@ -927,14 +947,18 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformReadOp(const substrait::Rel &so
 		scan = rel->Alias(name);
 	} else if (sget.has_virtual_table()) {
 		// We need to handle a virtual table as a LogicalExpressionGet
-		if (!sget.virtual_table().values().empty()) {
-			auto literal_values = sget.virtual_table().values();
+		if (!sget.virtual_table().expressions().empty()) {
```
nielspardon, line 954: **This condition is now identical to the `else if` just below** (`} else if (!sget.virtual_table().expressions().empty())` → `GetValuesExpression(...)`), so that second branch is now unreachable and `GetValuesExpression` becomes dead code — this was its only call site.

Before this PR the first branch tested `values()` and the `else if` tested `expressions()`, so DuckDB-produced virtual tables (which populate `expressions` via `TransformExpressionGet` → `TransformExpr` in to_substrait.cpp) fell through to `GetValuesExpression`, which supports non-literal entries. Now they hit this literal-only branch and throw `Non-literal expressions in virtual table values are not supported` for any non-literal field — a capability regression.

Since `GetValuesExpression` is a strict superset of this literal-only path, the cleanest fix is to drop this branch entirely and let `GetValuesExpression` handle all `expressions` cases — that removes the dead code *and* restores non-literal `VALUES` support in one change.

src/from_substrait.cpp
```diff
@@ -199,6 +199,17 @@ unique_ptr<ParsedExpression> SubstraitToDuckDB::TransformLiteralExpr(const subst
...
 
+static idx_t ExtractLiteralInteger(const substrait::Expression &expr, const char *context_desc) {
+	if (!expr.has_literal()) {
+		throw NotImplementedException("Non-literal expressions in %s are not supported", context_desc);
+	}
+	auto val = TransformLiteralToValue(expr.literal());
+	if (val.IsNull()) {
+		throw NotImplementedException("NULL expressions in %s are not supported", context_desc);
+	}
+	return val.GetValue<int64_t>();
```
nielspardon: `GetValue<int64_t>()` is returned as `idx_t` (unsigned), so a negative literal is silently converted to a huge value rather than rejected. For `offset` that's a latent bug (an offset can never legitimately be negative) — an explicit `< 0` check would fail fast.

Note this wraparound is also what silently reproduces the old `count == -1` "unlimited" sentinel (→ `idx_t` max). It works, but only implicitly — see the producer-side note on `count_expr`.

src/to_substrait.cpp
```diff
@@ -1113,8 +1122,16 @@ substrait::Rel *DuckDBToSubstrait::TransformLimit(LogicalOperator &dop) {
...
-	stopn->set_count(limit_val);
+	// Set offset expression
+	auto offset_expr = make_uniq<substrait::Expression>();
+	offset_expr->mutable_literal()->set_i64(static_cast<int64_t>(offset_val));
+	stopn->set_allocated_offset_expr(offset_expr.release());
+
+	// Set count expression
+	auto count_expr = make_uniq<substrait::Expression>();
+	count_expr->mutable_literal()->set_i64(static_cast<int64_t>(limit_val));
+	stopn->set_allocated_count_expr(count_expr.release());
```
nielspardon: When the limit is `UNSET`, `limit_val` is `-1` (set a few lines above), so this emits `count_expr = literal(i64: -1)`. On the consumer side that comes back as `-1` and only becomes "unlimited" via unsigned wraparound to `idx_t` max in `ExtractLiteralInteger`; the old explicit `count() == -1 ? Maximum` handling is gone.

Cleaner and more spec-faithful: simply don't set `count_expr` when the limit is unset (the consumer already guards with `has_count_expr()`), rather than emitting a literal `-1` count that external consumers won't interpret as "all rows".

src/from_substrait.cpp
```diff
@@ -927,14 +947,18 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformReadOp(const substrait::Rel &so
...
 			vector<vector<Value>> expression_rows;
 			for (auto &row : literal_values) {
 				auto values = row.fields();
 				vector<Value> expression_row;
 				for (const auto &value : values) {
-					expression_row.emplace_back(TransformLiteralToValue(value));
+					if (value.has_literal()) {
+						expression_row.emplace_back(TransformLiteralToValue(value.literal()));
+					} else {
+						throw NotImplementedException("Non-literal expressions in virtual table values are not supported");
```
nielspardon: Nit: message still says "virtual table values" though the field is now `expressions`. (Moot if this branch is removed per the comment above.)

nielspardon (commented):
Thanks for the 0.98.0 migration! The version bumps and the FetchRel expression-based migration look correct, and the new error-path test (`test_fetch_non_literal_expressions.test`) is well constructed — I verified the count/offset errors fire before input/table resolution.

One blocking issue plus a couple of round-trip semantics notes inline. The main one: the two virtual-table branches in `TransformReadOp` now share an identical condition, which makes `GetValuesExpression` unreachable and drops non-literal `VALUES` support.

nielspardon (commented):
Re-checked after the update — all my earlier feedback is addressed, and the fixes line up with the `FetchRel` spec. LGTM, thanks!

src/from_substrait.cpp
```diff
@@ -199,6 +199,21 @@ unique_ptr<ParsedExpression> SubstraitToDuckDB::TransformLiteralExpr(const subst
 	return make_uniq<ConstantExpression>(TransformLiteralToValue(sexpr.literal()));
 }
 
+static idx_t ExtractLiteralInteger(const substrait::Expression &expr, const char *context_desc) {
+	if (!expr.has_literal()) {
+		throw NotImplementedException("Non-literal expressions in %s are not supported", context_desc);
+	}
+	auto val = TransformLiteralToValue(expr.literal());
+	if (val.IsNull()) {
+		throw NotImplementedException("NULL expressions in %s are not supported", context_desc);
```
nielspardon, line 208: Optional / non-blocking: per the `FetchRel` spec a null-valued expression here isn't an error — for `count_expr` a null "signals that ALL records should be returned", and for `offset_expr` it "is treated as 0". This shared helper throws `NotImplementedException` for both.

Because the null semantics differ between count and offset, fully honoring the spec would mean handling null at the two call sites (count-null → `Maximum`, offset-null → `0`) rather than in the shared helper. It's very unusual for a producer to emit an explicit null limit, so this is fine to leave as-is — flagging only for spec completeness.

nielspardon (approved):
LGTM

PR #249, opened 2026-07-23
feat: add support for Substrait LateralJoinRel conversion (issue #243)

### LINKED ISSUE #243: Add LateralJoinRel Support to DuckDB Substrait Extension
# Add LateralJoinRel Support to DuckDB Substrait Extension

## Summary
Following the update to Substrait 0.96.0, implement support for converting DuckDB's `LOGICAL_LATERAL_JOIN` operator to/from the Substrait `LateralJoinRel` relation type. This enables proper round-trip conversion of correlated subqueries and lateral join operations.

## Background
- Substrait 0.96.0 introduced `LateralJoinRel` as a new relation type for representing lateral (correlated) joins
- DuckDB already has `LOGICAL_LATERAL_JOIN` in its query planner for handling correlated subqueries and LATERAL clauses
- The extension currently lacks conversion logic for this operator type, creating a gap in coverage for queries involving lateral joins

### To-Substrait Conversion (src/to_substrait.cpp)
1. **Add case handler for LOGICAL_LATERAL_JOIN** in the logical operator type switch statement
   - Create a `substrait::LateralJoinRel` protobuf message
   - Convert the left input relation
   - Convert the right input relation (correlated side)
   - Extract and convert the join condition expression
   - Extract and map correlated column references
   - Set appropriate join type semantics based on join condition

2. **Map DuckDB correlation information to Substrait**
   - Identify correlated columns from the right-side input
   - Build column reference mappings for the correlation
   - Ensure proper scope handling for correlated references

### From-Substrait Conversion (src/from_substrait.cpp)
1. **Add case handler for kLateral** in the Rel type switch statement
   - Extract left and right inputs from the `LateralJoinRel`
   - Convert both relations recursively
   - Extract the join condition
   - Convert to DuckDB's `LOGICAL_LATERAL_JOIN` operator
   - Restore correlated column information to the logical plan

2. **Handle correlation restoration**
   - Map Substrait correlated columns back to DuckDB ColumnRef objects
   - Ensure the right input's column scopes reference the left input correctly

## Testing
- Add unit tests for round-trip conversion of simple lateral joins
- Verify correlated subquery handling maintains correctness
- Test edge cases: multiple correlations, nested lateral joins, various join types (INNER/LEFT)

## Related Issues
- #222: Substrait 0.96.0 Support Update

## Files to Modify
- `src/to_substrait.cpp` - Add LOGICAL_LATERAL_JOIN case handler
- `src/from_substrait.cpp` - Add kLateral case handler  
- `test/` - Add test cases for lateral join conversion

## Notes
- The LateralJoinRel protobuf message is already available in the updated Substrait dependency
- This work should align with how DuckDB's internal lateral join execution model works
- Consider whether any hints or optimization markers should be preserved in the conversion

andrew-coleman:
Implement bidirectional conversion of Substrait LateralJoinRel to support LATERAL joins and correlated table functions. This complements the existing DELIM_JOIN/DELIM_GET support for decorrelated joins.

Changes:
- Add TransformLateralJoinOp() to convert Substrait LateralJoinRel to DuckDB JoinRelation with proper join type mapping (INNER, LEFT, SEMI, ANTI, MARK, SINGLE)
- Update TransformOp dispatcher to handle kLateralJoin case
- Add comprehensive test coverage in test_lateral_join.test for:
  - Explicit LATERAL keyword joins (INNER and LEFT)
  - LATERAL with aggregates
  - Multiple correlated columns

The implementation reuses existing join transformation patterns, ensuring consistency and maintainability. LateralJoinRel join types are restricted to Substrait-valid types (no RIGHT/OUTER), matching Substrait specification.

Resolves #243

Assisted by AI.

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -787,6 +787,39 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformJoinOp(const substrait::Rel &so
 	                                     djointype);
 }
 
+shared_ptr<Relation> SubstraitToDuckDB::TransformLateralJoinOp(const substrait::Rel &sop) {
+	auto &slateral = sop.lateral_join();
+
+	JoinType djointype;
+	switch (slateral.type()) {
+	case substrait::JoinRel::JoinType::JoinRel_JoinType_JOIN_TYPE_INNER:
+		djointype = JoinType::INNER;
+		break;
+	case substrait::JoinRel::JoinType::JoinRel_JoinType_JOIN_TYPE_LEFT:
+		djointype = JoinType::LEFT;
+		break;
+	case substrait::JoinRel::JoinType::JoinRel_JoinType_JOIN_TYPE_LEFT_SINGLE:
+		djointype = JoinType::SINGLE;
+		break;
+	case substrait::JoinRel::JoinType::JoinRel_JoinType_JOIN_TYPE_LEFT_SEMI:
+		djointype = JoinType::SEMI;
+		break;
+	case substrait::JoinRel::JoinType::JoinRel_JoinType_JOIN_TYPE_LEFT_ANTI:
+		djointype = JoinType::ANTI;
+		break;
+	case substrait::JoinRel::JoinType::JoinRel_JoinType_JOIN_TYPE_LEFT_MARK:
+		djointype = JoinType::MARK;
+		break;
+	default:
+		throw NotImplementedException("Unsupported LateralJoinRel join type");
+	}
+
+	unique_ptr<ParsedExpression> join_condition = TransformExpr(slateral.expression());
+	return make_shared_ptr<JoinRelation>(TransformOp(slateral.left())->Alias("left"),
+	                                     TransformOp(slateral.right())->Alias("right"), std::move(join_condition),
+	                                     djointype);
+}
+
 shared_ptr<Relation> SubstraitToDuckDB::TransformCrossProductOp(const substrait::Rel &sop) {
 	auto &sub_cross = sop.cross();
 
@@ -1506,6 +1539,8 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformOp(const substrait::Rel &sop,
 	switch (sop.rel_type_case()) {
 	case substrait::Rel::RelTypeCase::kJoin:
 		return TransformJoinOp(sop);
+	case substrait::Rel::RelTypeCase::kLateralJoin:
+		return TransformLateralJoinOp(sop);
 	case substrait::Rel::RelTypeCase::kCross:
 		return TransformCrossProductOp(sop);
 	case substrait::Rel::RelTypeCase::kFetch:
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -12,9 +12,11 @@
 #include "duckdb/common/shared_ptr.hpp"
 #include "duckdb/common/types.hpp"
 #include "duckdb/common/enums/set_operation_type.hpp"
+#include "duckdb/common/enums/joinref_type.hpp"
 #include "duckdb/planner/operator/logical_cteref.hpp"
 
 #include "duckdb/parser/expression/comparison_expression.hpp"
+#include "duckdb/parser/parsed_expression_iterator.hpp"
 
 #include "duckdb/main/client_data.hpp"
 #include "google/protobuf/unknown_field_set.h"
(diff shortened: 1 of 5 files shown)


nielspardon (commented):
Thanks for tackling this, @andrew-coleman — `LateralJoinRel` support is a real gap and the join-type mapping reads cleanly. Before merging I have a few concerns, the main one being that the new code path isn't actually exercised.

### 1. The new `TransformLateralJoinOp` is never reached by the tests (blocking)

All three test queries call `get_substrait('… LATERAL …')`. With `PRAGMA enable_verification` set, that *does* round-trip through the consumer (`VerifyBlobRoundtrip`/`VerifyJSONRoundtrip` → `SubstraitPlanToDuckDBRel` → `TransformOp`) — but the plan it round-trips never contains a `LateralJoinRel`. As the test comments note, `LATERAL` flattens to `LOGICAL_DELIM_JOIN`, and `to_substrait.cpp` lowers that to a `kJoin` + aggregates (`TransformComparisonJoin`), never `mutable_lateral_join()`. So `kLateralJoin` is never dispatched and `TransformLateralJoinOp` has zero coverage — the suite passes identically if the function body is deleted.

Since nothing in the extension *produces* a `LateralJoinRel`, the only way to cover the new code is to feed a hand-authored plan through `from_substrait_json(...)` and assert the result rows, as in `test/sql/test_from_substrait_basic.test`. That test would also surface the correctness issue below.

### 2. Correlated references aren't handled — silent wrong results (blocking)

A `LateralJoinRel`'s right input references left-row fields via `OuterReference` (a `FieldReference` `root_type`, pointing at the join's `RelCommon.rel_anchor`). But `TransformSelectionExpr` (`from_substrait.cpp:203-207`) only inspects `direct_reference`/`struct_field` and ignores `root_type`, so an outer reference is silently treated as an ordinary *local* positional field. A genuinely correlated lateral plan is therefore miscompiled to reference the wrong columns (or errors in the binder if the index is out of range) rather than failing loudly. `slateral.common()`/`rel_anchor` is also ignored.

(For what it's worth, building the `JoinRelation` with the default `JoinRefType::REGULAR` is fine — DuckDB detects laterality at bind time via `LateralBinder`, so no ref-type change is needed. The gap is purely the correlation decoding.)

Given this only works for the uncorrelated/degenerate case, I'd suggest either implementing `OuterReference` → correlated `ColumnRef` restoration, or explicitly throwing `NotImplementedException` when the right subtree carries outer references, so unsupported plans fail cleanly instead of returning wrong answers.

### 3. `post_join_filter` is silently dropped

`LateralJoinRel.post_join_filter` (field 5) is "semantically equivalent to a `FilterRel` directly above the join," but the handler never reads it. A plan that sets it would be silently mis-converted. At minimum, guard with a `NotImplementedException` if `slateral.has_post_join_filter()`. (Note `TransformJoinOp` has the same latent omission for `JoinRel.post_join_filter`, so this is a pre-existing pattern rather than new — but new code is a good place to start doing it right.)

### 4. Scope vs. #243 / PR description

The PR only touches `from_substrait.cpp` + the header (consumer direction). The description says "bidirectional conversion" and the test file title says "round-tripping," but there's no `to_substrait.cpp` producer for `LateralJoinRel`, which #243 explicitly lists under "To-Substrait Conversion" and "Files to Modify." Could you either add the producer side or reword the description to reflect that this is consumer-side ingestion only (and adjust the test title)?

### Minor

- For a comma-style `LATERAL` with no `ON` condition, `slateral.expression()` is unset and `TransformExpr` hits its `default:` branch with `REX_TYPE_NOT_SET` — worth guarding.
- The join-type `switch` duplicates `TransformJoinOp`'s; a shared `JoinRel::JoinType → JoinType` helper (with the lateral variant rejecting `RIGHT`/`OUTER`) would cut the duplication.
- The `default:` error message ("Unsupported LateralJoinRel join type") is less useful than `TransformJoinOp`'s, which names the offending type.

### Summary

The direction is right, but as-is the feature is untested and silently mishandles the correlation that's the whole point of a lateral join. Minimum to merge: (1) a `from_substrait_json` test that pushes a real `LateralJoinRel` through the new code, (2) either correlation support or a loud `NotImplementedException` for correlated inputs and `post_join_filter`, and (3) description/scope alignment with #243.

src/from_substrait.cpp
```diff
@@ -366,21 +368,124 @@ static idx_t ExtractLiteralInteger(const substrait::Expression &expr, const char
...
+		auto steps_out = outer_ref.steps_out();
+		// steps_out=0 means immediately enclosing scope (valid per Substrait spec)
+		// steps_out must be strictly less than lateral_scopes.size() to avoid underflow
+		if (steps_out >= lateral_scopes.size()) {
+			throw NotImplementedException(
+			    "OuterReference.steps_out %d exceeds the current LateralJoinRel nesting depth "
+			    "(%d); only outer references resolvable within tracked lateral scopes are supported",
+			    steps_out, (int)lateral_scopes.size());
+		}
+		scope = &lateral_scopes[lateral_scopes.size() - steps_out - 1];
```
nielspardon: **`steps_out` is handled off-by-one vs the Substrait spec.**

The proto documents `OuterReference.steps_out` as *"Number of subquery boundaries to traverse up for this field's reference. **Must be >= 1**"* — i.e. 1-based, with `steps_out=1` = the immediately enclosing scope. This code (and the comment on line 399) treats it as 0-based:

- `steps_out=0` (invalid per spec) is accepted as the innermost scope.
- For a single lateral join (`lateral_scopes.size()==1`), a spec-conforming `steps_out=1` hits `1 >= 1` and is **rejected** with "exceeds nesting depth" — a valid single-level correlation errors out.
- Nested cases resolve one scope too far out.

Suggested fix: reject `steps_out == 0` and index `lateral_scopes[size - steps_out]`. The only `stepsOut` test (case d1) checks the over-depth error; a valid `steps_out=1` case would lock this down.

src/from_substrait.cpp
```diff
@@ -742,51 +847,223 @@ OrderByNode SubstraitToDuckDB::TransformOrder(const substrait::SortField &sordf)
...
+					if (child && child->GetExpressionClass() == ExpressionClass::POSITIONAL_REFERENCE) {
+						auto &pos_ref = child->Cast<PositionalReferenceExpression>();
+						if (pos_ref.index > NumericLimits<idx_t>::Maximum() - left_column_count) {
+							throw InvalidInputException(
+							    "Field index adjustment overflow: index %llu + left_column_count %llu exceeds maximum",
+							    (unsigned long long)pos_ref.index, (unsigned long long)left_column_count);
+						}
+						pos_ref.index += left_column_count;
+					}
+				});
```
nielspardon, line 984: **This only descends one level, despite the "Recursively adjust" comment.**

`ParsedExpressionIterator::EnumerateChildren` visits *immediate* children only, and the inner callback never re-invokes `adjust_field_indices`, so only depth-0 (a bare positional ref) and depth-1 (a direct child) references get shifted by `left_column_count`. A right-side positional reference nested ≥2 levels deep — e.g. a correlated filter `t2.amount + 1 = t1.id`, where `+` is the direct child of `=` and the positional ref to `t2.amount` is a grandchild — keeps its unshifted (too-small) index and silently binds to the wrong (left-side) column. All current test conditions are flat two-arg comparisons, so it's uncaught.

Fix: have the inner callback recurse into `adjust_field_indices` for each child (true recursion), and correct the comment.

src/from_substrait.cpp
```diff
@@ -742,51 +847,223 @@ OrderByNode SubstraitToDuckDB::TransformOrder(const substrait::SortField &sordf)
...
+					    check_outer_refs(ifthen.then());
+				    }
+				    if (expr.if_then().has_else_()) {
+					    check_outer_refs(expr.if_then().else_());
+				    }
+			    }
+			    if (expr.has_cast()) {
+				    check_outer_refs(expr.cast().input());
+			    }
+			    // Note: additional nested expression types should be added as needed
```
nielspardon, line 1040: Minor: this walker only recurses into `scalar_function`/`if_then`/`cast`, so an `OuterReference` buried in e.g. a `singular_or_list` (`IN`) or `nested` expression won't be detected here. In practice it fails safe — once the scope is popped below, `TransformExpr` → `ResolveOuterReference` throws because `lateral_scopes` is empty — so you could either complete the walker or drop it and rely on that pop-then-throw path. As written, the hand-rolled proto walk is easy to leave subtly incomplete (as the trailing comment concedes).

nielspardon (changes requested):
Thanks for the thorough rework, @andrew-coleman — this resolves all four points from the first pass.

I found two correctness bugs in the new code (see inline), both in paths the tests don't cover. Neither affects the tested queries, but both would misbehave on valid inputs.

One nit: the PR description still says "Implement bidirectional conversion" and lists only the old `test_lateral_join.test`. Since the producer side is intentionally not implemented, please reword to consumer-side ingestion + the decorrelation rationale.

Overall very close — I'd like the two bugs fixed (each small) with a covering test, plus the description tweak, before merge.

src/from_substrait.cpp
nielspardon: Thanks, but this doesn't resolve the off-by-one. `steps_out >= size` → `steps_out > size - 1` is arithmetically identical for `size ≥ 1` (and `size ≥ 1` is guaranteed by the `lateral_scopes.empty()` guard above), so behavior is unchanged — the indexing and the `// steps_out=0 means immediately enclosing` comment are the same. `steps_out=0` (which the proto disallows — "Must be >= 1") is still accepted and maps to the innermost scope, and a spec-conforming `steps_out=1` on a single lateral join still throws (`1 > 0`).

Here is a 1-based fix (this also removes the `size - 1` unsigned-underflow footgun):

```suggestion
		// steps_out is 1-based per the Substrait spec ("Must be >= 1"); steps_out=1 refers to
		// the immediately enclosing scope. Reject 0 and anything beyond the current nesting depth.
		if (steps_out == 0 || steps_out > lateral_scopes.size()) {
			throw NotImplementedException(
			    "OuterReference.steps_out %d is out of range for the current LateralJoinRel "
			    "nesting depth (%d); it must satisfy 1 <= steps_out <= depth",
			    steps_out, (int)lateral_scopes.size());
		}
		scope = &lateral_scopes[lateral_scopes.size() - steps_out];
```

⚠️ Commit this **together** with the one-line change I left on `test/sql/test_from_substrait_lateral_join.test` — test (j) currently sends `"stepsOut": 0`, which this code now rejects. (You can add both to a batch and commit them in one go.)

test/sql/test_from_substrait_lateral_join.test
```diff
@@ -0,0 +1,841 @@
...
+                  },
+                  "namedTable": {"names": ["t2"]}
+                }
+              },
+              "condition": {
+                "scalarFunction": {
+                  "functionReference": 1,
+                  "arguments": [
+                    {"value": {"selection": {"directReference": {"structField": {"field": 0}}, "rootReference": {}}}},
+                    {"value": {"selection": {"directReference": {"structField": {"field": 0}}, "outerReference": {"stepsOut": 0}}}}
```
nielspardon: Pairs with the 1-based `steps_out` fix on `src/from_substrait.cpp` — send the spec-minimum value `1` (which now maps to the immediately-enclosing scope) instead of `0`:

```suggestion
                    {"value": {"selection": {"directReference": {"structField": {"field": 0}}, "outerReference": {"stepsOut": 1}}}}
```

Also worth updating the comment on line 686 ("steps_out can be 0 (immediately enclosing)") to say `1`, so the case documents the spec-conforming value — the header on line 685 already says "steps_out=1 is accepted".

src/from_substrait.cpp
nielspardon: The code fix here is correct now — capturing `[&]` and recursing via `adjust_field_indices(child)` descends all levels, thanks.

One caveat: the new test (k) doesn't actually exercise it. Its condition is `equal(rootReference field0, CAST(outerReference))` — the only node at depth ≥2 is the `OuterReference`, which resolves to a `ColumnRefExpression` (a leaf needing no shift), so the old shallow code produced identical indices and (k) passes with or without the fix. A genuine regression test needs a right-side positional reference nested ≥2 deep, e.g. `equal(add(rootReference field0, literal), outerReference)` (which also needs the arithmetic-functions extension registered in the plan).

nielspardon (approved):
Thanks for the quick turnaround, @andrew-coleman — both fixes from the last round are in and CI is green.

- `steps_out` is now correctly 1-based (rejects `0`, maps `steps_out=1` to the immediately-enclosing scope), and test (j) exercises the spec-conforming `"stepsOut": 1`.
- The correlated-filter index adjustment now recurses through all nesting levels.

Approving. One **non-blocking** nudge for now-or-later: test (k) doesn't actually exercise the recursion fix — its only depth-≥2 node is an `OuterReference`, which resolves to a `ColumnRefExpression` that needs no index shift, so it passes with or without the fix. A test with a *right-side positional reference* nested ≥2 deep — e.g. `equal(add(rootReference field0, literal), outerReference)` (needs the arithmetic-functions extension registered in the plan) — would actually guard the recursion against regressing. Up to you whether to add it here or in a follow-up.

Nice work seeing this through — the LateralJoinRel support looks good.

PR #254, opened 2026-07-30
add dialect yaml

andrew-coleman:
Provides a formal specification of the substrait
relations, types, expressions and functions supported by this extension.

This PR was assisted by AI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 The DuckDB Substrait Extension is a [DuckDB Community Extension](https://duckdb.org/community_extensions/) that provides [Substrait](https://substrait.io) support to [DuckDB](https://www.duckdb.org).
 With this extension, DuckDB can produce Substrait plans from DuckDB queries as well as consume Substrait plans and execute them with DuckDB.
-The most commonly used logical relations and functions are supported, including those generated from all of the TPC-H and TPC-DS queries. The supported features are formally specified in the Substrait dialect file [duckdb_dialect.yaml](duckdb_dialect.yaml).
+The most commonly used logical relations and functions are supported: Substrait plans can be produced for all of the TPC-H and TPC-DS queries. The supported features are formally specified in the Substrait dialect file [duckdb_dialect.yaml](duckdb_dialect.yaml).
 
 ## Building
 
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 The DuckDB Substrait Extension is a [DuckDB Community Extension](https://duckdb.org/community_extensions/) that provides [Substrait](https://substrait.io) support to [DuckDB](https://www.duckdb.org).
 With this extension, DuckDB can produce Substrait plans from DuckDB queries as well as consume Substrait plans and execute them with DuckDB.
+The most commonly used logical relations and functions are supported, including those generated from all of the TPC-H and TPC-DS queries. The supported features are formally specified in the Substrait dialect file [duckdb_dialect.yaml](duckdb_dialect.yaml).
 
 ## Building
 
--- a/duckdb_dialect.yaml
+++ b/duckdb_dialect.yaml
@@ -1,0 +1,1998 @@
+---
+name: "DuckDB Substrait Dialect"
+
+# ============================================================================
+# SUPPORTED DATA TYPES
+# ============================================================================
+supported_types:
+  - type: "BOOL"
+    system_metadata:
+      name: "BOOLEAN"
+      supported_as_column: true
+
+  - type: "I8"
+    system_metadata:
+      name: "TINYINT"
+      supported_as_column: true
+
+  - type: "I16"
+    system_metadata:
+      name: "SMALLINT"
+      supported_as_column: true
+
+  - type: "I32"
+    system_metadata:
+      name: "INTEGER"
+      supported_as_column: true
+
+  - type: "I64"
+    system_metadata:
+      name: "BIGINT"
+      supported_as_column: true
+
+  - type: "FP32"
+    system_metadata:
+      name: "FLOAT"
+      supported_as_column: true
+
+  - type: "FP64"
+    system_metadata:
+      name: "DOUBLE"
+      supported_as_column: true
+
+  - type: "DECIMAL"
+    system_metadata:
+      name: "DECIMAL"
+      supported_as_column: true
+    max_precision: 38
+    max_scale: 38
+
+  - type: "STRING"
+    system_metadata:
+      name: "VARCHAR"
+      supported_as_column: true
+
+  - type: "VARCHAR"
+    system_metadata:
+      name: "VARCHAR"
+      supported_as_column: true
+
+  - type: "FIXED_CHAR"
+    system_metadata:
+      name: "CHAR"
+      supported_as_column: true
+
+  - type: "BINARY"
+    system_metadata:
+      name: "BLOB"
+      supported_as_column: true
+
+  - type: "DATE"
+    system_metadata:
+      name: "DATE"
+      supported_as_column: true
+
+  - type: "PRECISION_TIME"
+    system_metadata:
+      name: "TIME"
+      supported_as_column: true
+    max_precision: 6
+
+  - type: "PRECISION_TIMESTAMP"
+    system_metadata:
+      name: "TIMESTAMP"
+      supported_as_column: true
+    max_precision: 9
+
+  - type: "PRECISION_TIMESTAMP_TZ"
+    system_metadata:
+      name: "TIMESTAMP_TZ"
+      supported_as_column: true
+    max_precision: 6
+
+  - type: "STRUCT"
+    system_metadata:
+      name: "STRUCT"
+      supported_as_column: true
+
+  - type: "LIST"
+    system_metadata:
+      name: "LIST"
+      supported_as_column: true
+
+  - type: "MAP"
+    system_metadata:
+      name: "MAP"
+      supported_as_column: true
+
+  - type: "UUID"
+    system_metadata:
+      name: "UUID"
+      supported_as_column: true
+
+# ============================================================================
+# SUPPORTED EXPRESSIONS
+# ============================================================================
+supported_expressions:
+  - "LITERAL"
+  - "SELECTION"
+  - "SCALAR_FUNCTION"
+  - "WINDOW_FUNCTION"
+  - "CAST"
+  - "IF_THEN"
+  - "SWITCH"
+  - expression: "SUBQUERY"
+    subquery_types:
+      - "SCALAR"
+      - "IN_PREDICATE"
+
+# ============================================================================
+# SUPPORTED RELATIONS
+# ============================================================================
+supported_relations:
+  - relation: "READ"
+    read_types:
+      - "NAMED_TABLE"
+  - "FILTER"
+  - "PROJECT"
+  - "AGGREGATE"
+  - "SORT"
+  - "FETCH"
+  - relation: "JOIN"
+    join_types:
+      - "INNER"
+      - "LEFT"
+      - "RIGHT"
+      - "OUTER"
+      - "LEFT_SEMI"
+      - "RIGHT_SEMI"
+      - "LEFT_ANTI"
+      - "RIGHT_ANTI"
+      - "LEFT_MARK"
+  - "CROSS"
+  - "LATERAL_JOIN"
+  - relation: "SET"
+    operations:
+      - "UNION_ALL"
+      - "UNION_DISTINCT"
+      - "INTERSECTION_PRIMARY"
+      - "MINUS_PRIMARY"
+  - "WRITE"
+  - "REFERENCE"
+
+# ============================================================================
+# DEPENDENCIES
+# ============================================================================
+dependencies:
+  arithmetic: "extension:io.substrait:functions_arithmetic"
+  arithmetic_decimal: "extension:io.substrait:functions_arithmetic_decimal"
+  comparison: "extension:io.substrait:functions_comparison"
+  datetime: "extension:io.substrait:functions_datetime"
+  geometry: "extension:io.substrait:functions_geometry"
+  string: "extension:io.substrait:functions_string"
+  aggregate_generic: "extension:io.substrait:functions_aggregate_generic"
+  aggregate_approx: "extension:io.substrait:functions_aggregate_approx"
+  rounding: "extension:io.substrait:functions_rounding"
+  rounding_decimal: "extension:io.substrait:functions_rounding_decimal"
+  boolean: "extension:io.substrait:functions_boolean"
+  list: "extension:io.substrait:functions_list"
+  logarithmic: "extension:io.substrait:functions_logarithmic"
+  set: "extension:io.substrait:functions_set"
+
+# ============================================================================
+# SCALAR FUNCTIONS
+# ============================================================================
+supported_scalar_functions:
+  # Arithmetic Functions - Basic Operations
+  - source: "arithmetic"
+    name: "add"
+    system_metadata:
+      name: "+"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i8"
+      - "i16_i16"
+      - "i32_i32"
+      - "i64_i64"
+      - "fp32_fp32"
+      - "fp64_fp64"
+
+  - source: "arithmetic"
+    name: "subtract"
+    system_metadata:
+      name: "-"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i8"
+      - "i16_i16"
+      - "i32_i32"
+      - "i64_i64"
+      - "fp32_fp32"
+      - "fp64_fp64"
+
+  - source: "arithmetic"
+    name: "multiply"
+    system_metadata:
+      name: "*"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i8"
+      - "i16_i16"
+      - "i32_i32"
+      - "i64_i64"
+      - "fp32_fp32"
+      - "fp64_fp64"
+
+  - source: "arithmetic"
+    name: "divide"
+    system_metadata:
+      name: "/"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i8"
+      - "i16_i16"
+      - "i32_i32"
+      - "i64_i64"
+      - "fp32_fp32"
+      - "fp64_fp64"
+
+  - source: "arithmetic"
+    name: "modulus"
+    system_metadata:
+      name: "%"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i8"
+      - "i16_i16"
+      - "i32_i32"
+      - "i64_i64"
+
+  - source: "arithmetic"
+    name: "negate"
+    system_metadata:
+      name: "-"
+      notation: "PREFIX"
+    supported_impls:
+      - "i8"
+      - "i16"
+      - "i32"
+      - "i64"
+      - "fp32"
+      - "fp64"
+
+  # Arithmetic Functions - Exponentiation & Power
+  - source: "arithmetic"
+    name: "sqrt"
+    system_metadata:
+      name: "sqrt"
+      notation: "FUNCTION"
+    supported_impls:
+      - "i64"
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "exp"
+    system_metadata:
+      name: "exp"
+      notation: "FUNCTION"
+    supported_impls:
+      - "i64"
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "power"
+    system_metadata:
+      name: "power"
+      notation: "FUNCTION"
+    supported_impls:
+      - "i64_i64"
+      - "fp32_fp32"
+      - "fp64_fp64"
+
+  - source: "arithmetic"
+    name: "factorial"
+    system_metadata:
+      name: "factorial"
+      notation: "FUNCTION"
+    supported_impls:
+      - "i64"
+
+  # Arithmetic Functions - Sign & Absolute Value
+  - source: "arithmetic"
+    name: "abs"
+    system_metadata:
+      name: "abs"
+      notation: "FUNCTION"
+    supported_impls:
+      - "i8"
+      - "i16"
+      - "i32"
+      - "i64"
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "sign"
+    system_metadata:
+      name: "sign"
+      notation: "FUNCTION"
+    supported_impls:
+      - "i8"
+      - "i16"
+      - "i32"
+      - "i64"
+      - "fp32"
+      - "fp64"
+
+  # Trigonometric Functions
+  - source: "arithmetic"
+    name: "sin"
+    system_metadata:
+      name: "sin"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "cos"
+    system_metadata:
+      name: "cos"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "tan"
+    system_metadata:
+      name: "tan"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "asin"
+    system_metadata:
+      name: "asin"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "acos"
+    system_metadata:
+      name: "acos"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "atan"
+    system_metadata:
+      name: "atan"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "atan2"
+    system_metadata:
+      name: "atan2"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32_fp32"
+      - "fp64_fp64"
+
+  - source: "arithmetic"
+    name: "degrees"
+    system_metadata:
+      name: "degrees"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "radians"
+    system_metadata:
+      name: "radians"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  # Hyperbolic Functions
+  - source: "arithmetic"
+    name: "sinh"
+    system_metadata:
+      name: "sinh"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "cosh"
+    system_metadata:
+      name: "cosh"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "tanh"
+    system_metadata:
+      name: "tanh"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "asinh"
+    system_metadata:
+      name: "asinh"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "acosh"
+    system_metadata:
+      name: "acosh"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "atanh"
+    system_metadata:
+      name: "atanh"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  # Bitwise Operations
+  - source: "arithmetic"
+    name: "bitwise_and"
+    system_metadata:
+      name: "&"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i8"
+      - "i16_i16"
+      - "i32_i32"
+      - "i64_i64"
+
+  - source: "arithmetic"
+    name: "bitwise_or"
+    system_metadata:
+      name: "|"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i8"
+      - "i16_i16"
+      - "i32_i32"
+      - "i64_i64"
+
+  - source: "arithmetic"
+    name: "bitwise_xor"
+    system_metadata:
+      name: "^"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i8"
+      - "i16_i16"
+      - "i32_i32"
+      - "i64_i64"
+
+  - source: "arithmetic"
+    name: "bitwise_not"
+    system_metadata:
+      name: "~"
+      notation: "PREFIX"
+    supported_impls:
+      - "i8"
+      - "i16"
+      - "i32"
+      - "i64"
+
+  - source: "arithmetic"
+    name: "shift_left"
+    system_metadata:
+      name: "<<"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i32"
+      - "i16_i32"
+      - "i32_i32"
+      - "i64_i32"
+
+  - source: "arithmetic"
+    name: "shift_right"
+    system_metadata:
+      name: ">>"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i32"
+      - "i16_i32"
+      - "i32_i32"
+      - "i64_i32"
+
+  - source: "arithmetic"
+    name: "shift_right_unsigned"
+    system_metadata:
+      name: ">>>"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i32"
+      - "i16_i32"
+      - "i32_i32"
+      - "i64_i32"
+
+  # Rounding Functions
+  - source: "rounding"
+    name: "ceil"
+    system_metadata:
+      name: "ceil"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "rounding"
+    name: "floor"
+    system_metadata:
+      name: "floor"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "rounding"
+    name: "round"
+    system_metadata:
+      name: "round"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp32_i32"
+      - "fp64"
+      - "fp64_i32"
+
+  # Comparison Functions
+  - source: "comparison"
+    name: "equal"
+    system_metadata:
+      name: "="
+      notation: "INFIX"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "not_equal"
+    system_metadata:
+      name: "!="
+      notation: "INFIX"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "lt"
+    system_metadata:
+      name: "<"
+      notation: "INFIX"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "lte"
+    system_metadata:
+      name: "<="
+      notation: "INFIX"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "gt"
+    system_metadata:
+      name: ">"
+      notation: "INFIX"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "gte"
+    system_metadata:
+      name: ">="
+      notation: "INFIX"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "is_null"
+    system_metadata:
+      name: "IS NULL"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any"
+
+  - source: "comparison"
+    name: "is_not_null"
+    system_metadata:
+      name: "IS NOT NULL"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any"
+
+  - source: "comparison"
+    name: "is_distinct_from"
+    system_metadata:
+      name: "IS DISTINCT FROM"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "is_not_distinct_from"
+    system_metadata:
+      name: "IS NOT DISTINCT FROM"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "is_nan"
+    system_metadata:
+      name: "is_nan"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "comparison"
+    name: "is_finite"
+    system_metadata:
+      name: "is_finite"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "comparison"
+    name: "is_infinite"
+    system_metadata:
+      name: "is_infinite"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp64"
+
+  - source: "comparison"
+    name: "is_true"
+    system_metadata:
+      name: "is_true"
+      notation: "FUNCTION"
+    supported_impls:
+      - "bool"
+
+  - source: "comparison"
+    name: "is_not_true"
+    system_metadata:
+      name: "is_not_true"
+      notation: "FUNCTION"
+    supported_impls:
+      - "bool"
+
+  - source: "comparison"
+    name: "is_false"
+    system_metadata:
+      name: "is_false"
+      notation: "FUNCTION"
+    supported_impls:
+      - "bool"
+
+  - source: "comparison"
+    name: "is_not_false"
+    system_metadata:
+      name: "is_not_false"
+      notation: "FUNCTION"
+    supported_impls:
+      - "bool"
+
+  - source: "comparison"
+    name: "between"
+    system_metadata:
+      name: "BETWEEN"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any_any"
+
+  - source: "comparison"
+    name: "coalesce"
+    system_metadata:
+      name: "coalesce"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any"
+    variadic:
+      min: 1
+
+  - source: "comparison"
+    name: "nullif"
+    system_metadata:
+      name: "nullif"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "least"
+    system_metadata:
+      name: "least"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any"
+    variadic:
+      min: 2
+
+  - source: "comparison"
+    name: "least_skip_null"
+    system_metadata:
+      name: "least_skip_null"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any"
+    variadic:
+      min: 2
+
+  - source: "comparison"
+    name: "greatest"
+    system_metadata:
+      name: "greatest"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any"
+    variadic:
+      min: 2
+
+  - source: "comparison"
+    name: "greatest_skip_null"
+    system_metadata:
+      name: "greatest_skip_null"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any"
+    variadic:
+      min: 2
+
+  # Boolean Functions
+  - source: "boolean"
+    name: "and"
+    system_metadata:
+      name: "AND"
+      notation: "INFIX"
+    supported_impls:
+      - "bool_bool"
+    variadic:
+      min: 2
+
+  - source: "boolean"
+    name: "or"
+    system_metadata:
+      name: "OR"
+      notation: "INFIX"
+    supported_impls:
+      - "bool_bool"
+    variadic:
+      min: 2
+
+  - source: "boolean"
+    name: "not"
+    system_metadata:
+      name: "NOT"
+      notation: "PREFIX"
+    supported_impls:
+      - "bool"
+
+  - source: "boolean"
+    name: "xor"
+    system_metadata:
+      name: "XOR"
+      notation: "INFIX"
+    supported_impls:
+      - "bool_bool"
+
+  - source: "boolean"
+    name: "and_not"
+    system_metadata:
+      name: "and_not"
+      notation: "INFIX"
+    supported_impls:
+      - "bool_bool"
+
+  # String Functions - Case Conversion
+  - source: "string"
+    name: "upper"
+    system_metadata:
+      name: "upper"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+      - "fixedchar"
+
+  - source: "string"
+    name: "lower"
+    system_metadata:
+      name: "lower"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+      - "fixedchar"
+
+  - source: "string"
+    name: "capitalize"
+    system_metadata:
+      name: "capitalize"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+
+  - source: "string"
+    name: "title"
+    system_metadata:
+      name: "title"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+
+  - source: "string"
+    name: "initcap"
+    system_metadata:
+      name: "initcap"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+
+  - source: "string"
+    name: "swapcase"
+    system_metadata:
+      name: "swapcase"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+
+  # String Functions - Length
+  - source: "string"
+    name: "char_length"
+    system_metadata:
+      name: "char_length"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+
+  - source: "string"
+    name: "bit_length"
+    system_metadata:
+      name: "bit_length"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+
+  - source: "string"
+    name: "octet_length"
+    system_metadata:
+      name: "octet_length"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+
+  # String Functions - Searching & Matching
+  - source: "string"
+    name: "like"
+    system_metadata:
+      name: "like"
+      notation: "INFIX"
+    supported_impls:
+      - "string_string"
+      - "varchar_varchar"
+
+  - source: "string"
+    name: "contains"
+    system_metadata:
+      name: "contains"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_string"
+      - "varchar_varchar"
+
+  - source: "string"
+    name: "starts_with"
+    system_metadata:
+      name: "starts_with"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_string"
+      - "varchar_varchar"
+
+  - source: "string"
+    name: "ends_with"
+    system_metadata:
+      name: "ends_with"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_string"
+      - "varchar_varchar"
+
+  - source: "string"
+    name: "strpos"
+    system_metadata:
+      name: "strpos"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_string"
+      - "varchar_varchar"
+
+  - source: "string"
+    name: "substring"
+    system_metadata:
+      name: "substring"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_i32"
+      - "string_i32_i32"
+      - "varchar_i32"
+      - "varchar_i32_i32"
+
+  - source: "string"
+    name: "left"
+    system_metadata:
+      name: "left"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_i32"
+      - "varchar_i32"
+
+  - source: "string"
+    name: "right"
+    system_metadata:
+      name: "right"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_i32"
+      - "varchar_i32"
+
+  # String Functions - Regular Expressions
+  - source: "string"
+    name: "regexp_match_substring"
+    system_metadata:
+      name: "regexp_match_substring"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_string"
+      - "varchar_varchar"
+
+  - source: "string"
+    name: "regexp_match_substring_all"
+    system_metadata:
+      name: "regexp_match_substring_all"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_string"
+      - "varchar_varchar"
+
+  - source: "string"
+    name: "regexp_strpos"
+    system_metadata:
+      name: "regexp_strpos"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_string"
+      - "varchar_varchar"
+
+  - source: "string"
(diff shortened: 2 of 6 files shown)


vcpkg.json
```diff
@@ -1,7 +1,10 @@
 {
     "dependencies": [
         "protobuf",
-        "substrait-protobuf"
+        "substrait-protobuf",
+        "nlohmann-json",
+        "json-schema-validator",
+        "yaml-cpp"
     ],
```
nielspardon, line 9: The schema and the extension YAMLs both ship in the `substrait-extensions` port, which is already whitelisted in `vcpkg-configuration.json`.

```suggestion
    "dependencies": [
        "protobuf",
        "substrait-protobuf",
        "substrait-extensions",
        "nlohmann-json",
        "json-schema-validator",
        "yaml-cpp"
    ],
```

(`nlohmann-json` is already transitive — `json-schema-validator` 2.4.0 depends on it and `nlohmann_json_schema_validator::validator` carries `INTERFACE_LINK_LIBRARIES "nlohmann_json::nlohmann_json"` — so listing it is a style choice. I'd keep it, since `tojson.hpp` includes `<nlohmann/json.hpp>` directly.)

---

vcpkg.json
```diff
@@ -1,7 +1,10 @@
...
     "builtin-baseline": "84bab45d415d22042bd0b9081aea57f362da3f35",
     "overrides": [
```
nielspardon, line 11: Pinning it to `0.98.0` matches the tag `scripts/generate_custom_functions.py:16` already uses, so the dialect is validated against the same definitions the generated function registry came from.

```suggestion
    "overrides": [
        {
            "name": "substrait-extensions",
            "version": "0.98.0"
        },
```

---

test/c/CMakeLists.txt
```diff
@@ -4,14 +4,56 @@ project(${TARGET_NAME})
...
+  configure_file(
+    "${DIALECT_SCHEMA_SOURCE}/dialect_schema.yaml"
+    "${CMAKE_BINARY_DIR}/dialect_schema.yaml"
+    COPYONLY
+  )
+else()
+  message(FATAL_ERROR "Could not find dialect_schema.yaml in substrait-protobuf package. "
+    "Ensure substrait-protobuf is installed via vcpkg and your build directory contains vcpkg_installed/. "
+    "Set VCPKG_TARGET_TRIPLET if using multiple platform triplets.")
+endif()
```
nielspardon: `substrait-protobuf` never installs `dialect_schema.yaml`. At the pinned 0.98.0 the port installs `include/substrait/*.pb.h`, the vendored `.proto` tree into `share/substrait/proto/`, and the CMake config; `share/substrait-protobuf/` holds only vcpkg's own `copyright`, `vcpkg.spdx.json` and `vcpkg_abi_info.txt`. The schema lives at the *repo root's* `substrait/text/`, which is never part of the install — so the first branch can't ever hit.

That leaves the buildtree glob, which reaches into vcpkg's private scratch space:

- it needs `VCPKG_ROOT` set;
- it needs the buildtree to still exist — a binary-cache hit means the port is never extracted, so there is no buildtree at all, and `FATAL_ERROR` then aborts configuration of the whole extension (`SKIP_SUBSTRAIT_C_TESTS` is set nowhere in CI);
- `list(GET DIALECT_SCHEMA_SEARCH 0 ...)` picks whichever tree sorts first. I have **ten** extracted `src/*` trees for this port locally, holding **three distinct** `dialect_schema.yaml` versions — so the schema actually validated against is not necessarily the pinned one.

`substrait-extensions` solves this properly. Verified by installing it at 0.98.0:

```
share/substrait/text/dialect_schema.yaml
share/substrait/extensions/*.yaml          # identical to the cpp/substrait-extensions/v0.98.0 tag
SubstraitExtensions_DATA_DIR = <prefix>/share/substrait
```

```suggestion
include_directories(include)
include_directories(../../duckdb/src/include)
include_directories(../../duckdb/test/include)
include_directories(../../duckdb/third_party/catch)
find_package(nlohmann_json_schema_validator REQUIRED)
find_package(yaml-cpp REQUIRED)

# The dialect schema and the extension definitions both ship in the
# substrait-extensions port (data-only). SubstraitExtensions_DATA_DIR points at
# the installed data root: text/dialect_schema.yaml and extensions/*.yaml. It is
# pinned to the same version as SUBSTRAIT_EXTENSIONS_TAG in
# scripts/generate_custom_functions.py, so the dialect is checked against the
# definitions the generated function registry was built from.
find_package(SubstraitExtensions CONFIG REQUIRED)
```

This also drops the added `find_package(SubstraitProtobuf CONFIG REQUIRED)`: nothing under `test/c` includes a `.pb.h`, and `test_substrait_exe` gets whatever it needs through `substrait_extension`, which is how it works on `main` today.

The `target_compile_definitions` on line 56 needs updating to match — see the next comment.

---

test/c/CMakeLists.txt
```diff
@@ -4,14 +4,56 @@ project(${TARGET_NAME})
...
+
+set(ALL_SOURCES test_substrait_c_api.cpp test_substrait_c_utils.cpp test_projection.cpp test_base_schema_projection.cpp test_root_names.cpp test_dialect_validation.cpp)
 
 
 add_library_unity(test_substrait OBJECT ${ALL_SOURCES})
+target_link_libraries(test_substrait nlohmann_json::nlohmann_json nlohmann_json_schema_validator::validator yaml-cpp::yaml-cpp)
+target_compile_definitions(test_substrait PRIVATE "DIALECT_SCHEMA_DIR=\"${CMAKE_BINARY_DIR}\"")
```
nielspardon: With the schema read from `SubstraitExtensions_DATA_DIR` there's nothing to copy into the build dir, so `DIALECT_SCHEMA_DIR` becomes the wrong shape. Passing both paths also fixes the test's dependence on the working directory (see C4):

```suggestion
target_compile_definitions(test_substrait PRIVATE
    "SUBSTRAIT_DATA_DIR=\"${SubstraitExtensions_DATA_DIR}\""
    "DIALECT_YAML=\"${CMAKE_CURRENT_SOURCE_DIR}/../../duckdb_dialect.yaml\"")
```

For the record, I went looking for a link problem here and there isn't one: `set(ALL_OBJECT_FILES ... PARENT_SCOPE)` on lines 57-59 is inert for an out-of-tree extension — its parent is this repo's top-level `CMakeLists.txt`, which never forwards the variable, so `test/c` objects are not linked into DuckDB's `unittest` (`build test/unittest:` in a configured `build.ninja` contains only `generated_extension_loader.cpp.o`, `ub_test_sqlite.cpp.o` and `unittest.cpp.o`). The dialect test therefore only ever runs via `test_substrait_exe`, which `make release_c_unit_test` enables and `c_tests.yml` runs — so CI does cover it. The extra libraries on line 66 are needed simply because `$<TARGET_OBJECTS:...>` carries no usage requirements, which the PR already handles correctly.

---

test/c/test_dialect_validation.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+TEST_CASE("Dialect YAML conforms to schema", "[dialect-validation]") {
+	json dialect = loadyaml("duckdb_dialect.yaml");
+	json schema = loadyaml(std::string(DIALECT_SCHEMA_DIR) + "/dialect_schema.yaml");
+
+    json_validator validator;
+	REQUIRE_NOTHROW(validator.set_root_schema(schema));
+	REQUIRE_NOTHROW(validator.validate(dialect));
+}
+
```
nielspardon: Two things:

- `"duckdb_dialect.yaml"` is relative, so the test only passes because CI happens to run from the repo root. The schema path already gets a compile-time define; the dialect path should too.
- `REQUIRE_NOTHROW` does surface the validator's message — Catch2 prints `due to unexpected exception with message:` followed by `e.what()`, and json-schema-validator's `what()` carries the JSON pointer (`At /supported_types/1/type of "NOPE" - instance not found in required enum`). What it *doesn't* say is which two files were compared, since both paths are literals inside the assertion. Catching and passing `e.what()` to `FAIL` is worth it for that, and it keeps the assertion count honest.
- The two "file exists" cases are then subsumed: `loadyaml` throws `YAML::BadFile` if either path is wrong, and that now names the path.

```suggestion
#include "catch.hpp"
#include <string>
#include <nlohmann/json-schema.hpp>
#include <tojson.hpp>

using nlohmann::json;
using nlohmann::json_schema::json_validator;

// Both paths are baked in at configure time (see test/c/CMakeLists.txt): the
// dialect lives in the source tree and the schema in the installed
// substrait-extensions data dir, so neither depends on the working directory.
static const std::string DIALECT_PATH = DIALECT_YAML;
static const std::string SCHEMA_PATH = std::string(SUBSTRAIT_DATA_DIR) + "/text/dialect_schema.yaml";

TEST_CASE("Dialect YAML conforms to schema", "[dialect-validation]") {
	json dialect = tojson::loadyaml(DIALECT_PATH);
	json schema = tojson::loadyaml(SCHEMA_PATH);

	json_validator validator;
	validator.set_root_schema(schema);
	try {
		validator.validate(dialect);
	} catch (const std::exception &e) {
		// Name the files: the validator's own message gives the JSON pointer but
		// not which dialect was checked against which schema.
		FAIL(DIALECT_PATH << " does not conform to " << SCHEMA_PATH << ": " << e.what());
	}
}
```

Worth confirming separately: `nlohmann/json-schema-validator` targets **draft-07**, while `dialect_schema.yaml` declares `$schema: https://json-schema.org/draft/2020-12/schema`. It validates today, but if the library ignores rather than rejects an unknown `$schema`, keywords it doesn't implement pass silently. A deliberately-invalid fixture that must fail would pin that down.

---

test/c/test_dialect_validation.cpp
```diff
@@ -0,0 +1,31 @@
...
+	REQUIRE(file.good());
+}
+
+TEST_CASE("Dialect schema file exists", "[dialect-validation]") {
+	const std::string schemaPath = std::string(DIALECT_SCHEMA_DIR) + "/dialect_schema.yaml";
+	std::ifstream file(schemaPath);
+	REQUIRE(file.good());
+}
+
+TEST_CASE("Dialect YAML conforms to schema", "[dialect-validation]") {
```
nielspardon, line 15: Schema conformance only proves the file is *shaped* right; it can't see that a `supported_impls` string names no real implementation. That's where the errors are — 193 of 388 declared impls don't resolve against `functions_*.yaml` at v0.98.0.

Once `SubstraitExtensions_DATA_DIR` is available you get `extensions/*.yaml` for free, so this is cheap to add. For every `(source, name, impl)`: resolve `dependencies[source]` to the extension file by `urn`, find `name` under the matching function kind, and require `impl` to equal the argument signature of one of its `impls` — short names per <https://substrait.io/extensions/#function-signature-compound-names>, `req` for enumeration args, the variadic arg counted once, `""` for zero-arg.

`src/custom_extensions.cpp:214-236` already carries this table, but sharing it takes two changes rather than one: `TypeShortName` is `static` with no declaration in `src/include/`, so it has to be lifted into a header before a test target can call it at all; and its keys are the protobuf `Type.kind` field names (`fixed_char`, `fixed_binary`) whereas the YAMLs spell those `fixedchar<L1>` / `fixedbinary` and sometimes upper-case the type (`DECIMAL<P, S>`), so it needs YAML-side keys and case-folding too.

A companion check on `system_metadata.name` against `duckdb_functions()` would be useful too, but it needs an explicit allowlist, and not a small one. Every non-`FUNCTION` entry has to be exempt — 31 `INFIX` and 3 `PREFIX` in the current file, and at least `boolean.not` (`"NOT"`, `PREFIX`) is absent from the catalog. So are `BETWEEN` and `coalesce`, which are `FUNCTION`-notation and written like calls but aren't catalog rows either. The check is still worth having; it just can't be a plain set-membership test.

---

test/c/include/tojson.hpp
```diff
@@ -0,0 +1,86 @@
...
+	// string tag will be !
+	if (node.Tag() != "!") {
+		if (YAML::convert<int>::decode(node, i)) return i;
+		if (YAML::convert<double>::decode(node, d)) return d;
+		if (YAML::convert<bool>::decode(node, b)) return b;
+	}
+	if (YAML::convert<std::string>::decode(node, s)) return s;
+	
+	return nullptr;
+}
```
nielspardon, line 39: `int` truncates: any integer above `INT_MAX` fails the `int` decode, falls through to `double`, and then fails a `type: integer` check downstream. Harmless for these two files today (the largest literal is `2147483647`), but it costs nothing to avoid. `yaml-cpp` specialises `convert` for both `long` and `long long`, so `int64_t` resolves on any platform.

Also `detail::repr()` and `#include <fstream>` are dead in this stripped copy, and line 38 has trailing whitespace.

```suggestion
inline nlohmann::json parse_scalar(const YAML::Node &node) {
	int64_t i;
	double d;
	bool b;
	std::string s;

	// string tag will be !
	if (node.Tag() != "!") {
		if (YAML::convert<int64_t>::decode(node, i)) return i;
		if (YAML::convert<double>::decode(node, d)) return d;
		if (YAML::convert<bool>::decode(node, b)) return b;
	}
	if (YAML::convert<std::string>::decode(node, s)) return s;

	return nullptr;
}
```

Vendoring is fine — MIT, attributed inline, repo LICENSE is MIT — but this is the first vendored file here, so it sets the precedent. Worth a note in the README or a `third_party/` convention.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+
+  - type: "STRING"
+    system_metadata:
+      name: "VARCHAR"
+      supported_as_column: true
+
+  - type: "VARCHAR"
+    system_metadata:
+      name: "VARCHAR"
+      supported_as_column: true
```
nielspardon, line 60: Both entries map to DuckDB `VARCHAR`, which hides an asymmetry: the producer only ever emits `string` (`to_substrait.cpp:1930`), while the consumer accepts both `kString` and `kVarchar` (`from_substrait.cpp:650-652`). The format can't express "consume-only", but `metadata` can:

```suggestion
  - type: "STRING"
    system_metadata:
      name: "VARCHAR"
      supported_as_column: true

  - type: "VARCHAR"
    metadata:
      direction: "CONSUME_ONLY"  # producer always emits STRING for DuckDB VARCHAR
    system_metadata:
      name: "VARCHAR"
      supported_as_column: true
```

Something similar applies to the precision bounds a few lines down, though it's narrower than it first looks. In the *column-type* mapping, `PRECISION_TIMESTAMP` accepts only {0, 3, 6, 9} and `PRECISION_TIMESTAMP_TZ` requires exactly 6 — `from_substrait.cpp:659-682` throws otherwise. The *literal* path is looser: both go through `ScaleToMicros`, which accepts any precision in 0-9 and rescales (`from_substrait.cpp:138-148`, `245-258`). So `max_precision: 9` / `max_precision: 6` describe the literal path reasonably but overstate the column path, and `max_precision` alone can't express "{0, 3, 6, 9} as a column, 0-9 as a literal". A `metadata` note is probably the only way to say it.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+
+  - type: "FIXED_CHAR"
+    system_metadata:
+      name: "CHAR"
+      supported_as_column: true
```
nielspardon, line 70: `FIXED_CHAR` isn't supported as a *column* type: `s_type.mutable_fixed_char()` appears nowhere in `to_substrait.cpp`, and `SubstraitToDuckType` has no `KindCase::kFixedChar`. DuckDB's `CHAR` is a plain alias for `VARCHAR`, so there's no fixed-length char type for it to mean.

The entry should stay, though — `supported_as_column: false` is the fix, not removal. The literal path does work (`from_substrait.cpp:235-236`), the schema says types listed here "are always supported as literals", and C18 needs `FIXED_CHAR` present because `upper`/`lower` genuinely have `fchar` impls upstream. Dropping it while declaring an `fchar` argument would be the same inconsistency C26 raises about `u!geometry`.

The interval types have the opposite problem — supported, but absent. Because `supported_types` is a closed list, their absence declares DuckDB's `INTERVAL` unsupported while the datetime functions below take `interval_day` operands. All three are needed, for different reasons, and they differ in shape:

- `INTERVAL_DAY` is a full column type (`to_substrait.cpp:1908` maps `LogicalTypeId::INTERVAL` to `interval_day`; `from_substrait.cpp:709` maps it back). This one the file already contradicts itself about — `datetime.add`/`subtract`/`multiply` declare `interval_day` operands today.
- `INTERVAL_YEAR` is literal-only: `mutable_interval_year` and `KindCase::kIntervalYear` appear nowhere, but the literal path works both ways (`to_substrait.cpp:207-209`, `from_substrait.cpp:217-219`). Nothing in the file references it *yet* — it becomes necessary once `add`/`subtract` gain their `iyear` signatures (C23), which is why it's here.
- `INTERVAL_COMPOUND` is also literal-only and also absent: `from_substrait.cpp:343-358` turns an `interval_compound` literal into a DuckDB `INTERVAL`.

`supported_as_column: false` is what the schema offers for the literal-only pair.

```suggestion
  - type: "FIXED_CHAR"
    system_metadata:
      name: "CHAR"
      supported_as_column: false

  - type: "INTERVAL_DAY"
    system_metadata:
      name: "INTERVAL"
      supported_as_column: true
    max_precision: 6

  - type: "INTERVAL_YEAR"
    system_metadata:
      name: "INTERVAL"
      supported_as_column: false

  - type: "INTERVAL_COMPOUND"
    system_metadata:
      name: "INTERVAL"
      supported_as_column: false
    max_precision: 6
```

`PRECISION_TIME` (lines 75-79) is a subtler case and I'd leave it as-is with a note: the *literal* producer does emit `precision_time` (`to_substrait.cpp:189`, `TransformTime`), but the *column-type* mapping routes DuckDB `TIME` to `precision_timestamp` (`to_substrait.cpp:1884-1889`). So the two halves of the producer disagree with each other — which looks like a bug in the extension rather than in this file, but it does mean `PRECISION_TIME` with `supported_as_column: true` overstates what a `TIME` column round-trips as.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+  - "SELECTION"
+  - "SCALAR_FUNCTION"
+  - "WINDOW_FUNCTION"
+  - "CAST"
+  - "IF_THEN"
+  - "SWITCH"
+  - expression: "SUBQUERY"
+    subquery_types:
+      - "SCALAR"
+      - "IN_PREDICATE"
```
nielspardon: Three claimed-but-absent, two supported-but-omitted. Since the list is closed, the omissions actively declare working features unsupported.

Absent:
- `SWITCH` — no `switch_expression` anywhere in `src/`.
- `WINDOW_FUNCTION` — as an *expression* it is never produced or consumed. Window functions travel via `ConsistentPartitionWindowRel`, which is itself missing from `supported_relations` (see C10).
- `SUBQUERY` — `case kSubquery:` falls straight through to `default: throw NotImplementedException` (`from_substrait.cpp:820-823`), and `to_substrait.cpp` never emits one. This is the largest of the three, since it's the only entry carrying a `subquery_types` list.

Supported both ways, omitted:
- `SINGULAR_OR_LIST` — produced at `to_substrait.cpp:522,825`, consumed at `from_substrait.cpp:816`.
- `NESTED` — produced at `to_substrait.cpp:327,336,345`, consumed at `from_substrait.cpp:818`.

```suggestion
supported_expressions:
  - "LITERAL"
  - "SELECTION"
  - "SCALAR_FUNCTION"
  - "CAST"
  - "IF_THEN"
  - "SINGULAR_OR_LIST"
  - expression: "NESTED"
    nested_types:
      - "STRUCT"
      - "LIST"
      - "MAP"
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+  - "CROSS"
+  - "LATERAL_JOIN"
+  - relation: "SET"
+    operations:
+      - "UNION_ALL"
+      - "UNION_DISTINCT"
+      - "INTERSECTION_PRIMARY"
+      - "MINUS_PRIMARY"
+  - "WRITE"
+  - "REFERENCE"
```
nielspardon, line 183: - `read_types: [NAMED_TABLE]` omits `VIRTUAL_TABLE` (`to_substrait.cpp:2067,2081,2280`; `from_substrait.cpp:1390`) and `LOCAL_FILES` (`to_substrait.cpp:2039`; `from_substrait.cpp:1359`).
- `join_types` omits `LEFT_SINGLE`, which is emitted (`to_substrait.cpp:1334`) and consumed (`from_substrait.cpp:872`).
- `SET.UNION_DISTINCT` — `UNION_DISTINCT` appears nowhere in `src/`. The producer emits `UNION_ALL` and wraps distinct in an aggregate (`to_substrait.cpp:2348`); the consumer handles only `UNION_ALL`, `MINUS_PRIMARY`, `INTERSECTION_PRIMARY` (`from_substrait.cpp:1689-1695`).
- `CONSISTENT_PARTITION_WINDOW` is produced (`to_substrait.cpp:1591`) and consumed (`from_substrait.cpp:1811` → `TransformWindowOp`) but omitted. Worth flagging hardest: the file declares 11 window functions but not the relation that carries them.

```suggestion
supported_relations:
  - relation: "READ"
    read_types:
      - "NAMED_TABLE"
      - "VIRTUAL_TABLE"
      - "LOCAL_FILES"
  - "FILTER"
  - "PROJECT"
  - "AGGREGATE"
  - "SORT"
  - "FETCH"
  - relation: "JOIN"
    join_types:
      - "INNER"
      - "LEFT"
      - "RIGHT"
      - "OUTER"
      - "LEFT_SEMI"
      - "RIGHT_SEMI"
      - "LEFT_ANTI"
      - "RIGHT_ANTI"
      - "LEFT_SINGLE"
      - "LEFT_MARK"
  - "CROSS"
  - "LATERAL_JOIN"
  - "CONSISTENT_PARTITION_WINDOW"
  - relation: "SET"
    operations:
      - "UNION_ALL"
      - "INTERSECTION_PRIMARY"
      - "MINUS_PRIMARY"
  - "WRITE"
  - "REFERENCE"
```

For the record, `HASH_JOIN`, `MERGE_JOIN`, `NESTED_LOOP_JOIN`, `EXCHANGE`, `EXPAND`, `UPDATE`, `DDL` and the three extension rels are **correctly absent** — I initially misread `GetCommon()` (`from_substrait.cpp:1103-1150`) as the consumer dispatch. It isn't; it only returns the `RelCommon` submessage for emit mapping and is reached from one call site. The real dispatch, `TransformOp` (`from_substrait.cpp:1790-1822`), accepts thirteen rel types and throws `NotImplementedException` for everything else, and `kUpdate`/`kDdl` fall into `GetCommon`'s own throwing `default:`.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+  - source: "arithmetic"
+    name: "shift_right_unsigned"
+    system_metadata:
+      name: ">>>"
+      notation: "INFIX"
+    supported_impls:
+      - "i8_i32"
+      - "i16_i32"
+      - "i32_i32"
+      - "i64_i32"
```
nielspardon, line 548: `functions_arithmetic.yaml` declares only `i32_i32` and `i64_i32` for all three; `i8_i32` and `i16_i32` don't exist, so a consumer resolving them fails.

```suggestion
  - source: "arithmetic"
    name: "shift_left"
    system_metadata:
      name: "<<"
      notation: "INFIX"
    supported_impls:
      - "i32_i32"
      - "i64_i32"

  - source: "arithmetic"
    name: "shift_right"
    system_metadata:
      name: ">>"
      notation: "INFIX"
    supported_impls:
      - "i32_i32"
      - "i64_i32"

  - source: "arithmetic"
    name: "shift_right_unsigned"
    system_metadata:
      name: ">>>"
      notation: "INFIX"
    supported_impls:
      - "i32_i32"
      - "i64_i32"
```

DuckDB has no `>>>` operator, so I'd drop `shift_right_unsigned` outright rather than fix its impls — your call on whether the entry is aspirational.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+  - source: "rounding"
+    name: "round"
+    system_metadata:
+      name: "round"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp32"
+      - "fp32_i32"
+      - "fp64"
+      - "fp64_i32"
```
nielspardon, line 576: Every `round` impl in `functions_rounding.yaml` takes two arguments; there is no single-argument form. (There are also `i8`/`i16`/`i32`/`i64` variants, which DuckDB supports too if you want them.)

```suggestion
    supported_impls:
      - "fp32_i32"
      - "fp64_i32"
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+      name: ">="
+      notation: "INFIX"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "is_null"
+    system_metadata:
+      name: "IS NULL"
+      notation: "FUNCTION"
```
nielspardon: `IS NULL` is syntax, not a catalog function — `duckdb_functions()` has no entry for it, and the same goes for `IS NOT NULL` (634). Postfix is the right notation for both:

```suggestion
      name: "IS NULL"
      notation: "POSTFIX"
```

`IS DISTINCT FROM` (642) and `IS NOT DISTINCT FROM` (650) are also absent from the catalog, but they're binary — `a IS DISTINCT FROM b`, declared over `any_any` — so `POSTFIX` ("a b +") doesn't describe them either. `INFIX` is the closest fit the schema offers. `BETWEEN` (717) is ternary and fits none of the four; a `metadata` note may be the only honest option.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+    system_metadata:
+      name: "IS NOT DISTINCT FROM"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any"
+
+  - source: "comparison"
+    name: "is_nan"
+    system_metadata:
+      name: "is_nan"
```
nielspardon: DuckDB spells this `isnan` (no underscore). Likewise `is_finite` → `isfinite` (667) and `is_infinite` → `isinf` (676) — none of the three declared names are in `duckdb_functions()` on 1.5.5.

```suggestion
      name: "isnan"
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+
+  - source: "comparison"
+    name: "coalesce"
+    system_metadata:
+      name: "coalesce"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any"
+    variadic:
+      min: 1
```
nielspardon: Upstream declares `variadic: {min: 2}`. A dialect may *narrow* an upstream bound (which is what `and`/`or` below do, going from upstream's `min: 0` to `min: 2`), but it can't widen it — `min: 1` claims an arity the extension doesn't define.

```suggestion
      min: 2
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+
+  - source: "comparison"
+    name: "greatest_skip_null"
+    system_metadata:
+      name: "greatest_skip_null"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any_any"
+    variadic:
+      min: 2
```
nielspardon: This contradicts a fix on this branch (733af82). `to_substrait.cpp:51-56` and `from_substrait.cpp:48-69` both say, with comments: DuckDB `least`/`greatest` map to Substrait `least_skip_null`/`greatest_skip_null`, and Substrait's null-propagating `least`/`greatest` are *deliberately absent* because DuckDB has no function with those semantics.

So the two plain entries are false positives, and the `_skip_null` pair carry the wrong DuckDB name — `least_skip_null` and `greatest_skip_null` are not DuckDB functions. The impls are wrong too: these are variadic over a single `any1`, so the signature is `any`, not `any_any`.

```suggestion
  - source: "comparison"
    name: "least_skip_null"
    system_metadata:
      name: "least"
      notation: "FUNCTION"
    supported_impls:
      - "any"
    variadic:
      min: 2

  - source: "comparison"
    name: "greatest_skip_null"
    system_metadata:
      name: "greatest"
      notation: "FUNCTION"
    supported_impls:
      - "any"
    variadic:
      min: 2
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+
+  - source: "boolean"
+    name: "or"
+    system_metadata:
+      name: "OR"
+      notation: "INFIX"
+    supported_impls:
+      - "bool_bool"
+    variadic:
+      min: 2
```
nielspardon, line 783: Variadic arguments appear **once** in a signature, so `and(boolean...)` is `and:bool`, not `bool_bool` — the spec's own reference dialect (`dialects/tests/functions_test.yaml`) has exactly this case.

```suggestion
  # Boolean Functions
  - source: "boolean"
    name: "and"
    system_metadata:
      name: "AND"
      notation: "INFIX"
    supported_impls:
      - "bool"
    variadic:
      min: 2

  - source: "boolean"
    name: "or"
    system_metadata:
      name: "OR"
      notation: "INFIX"
    supported_impls:
      - "bool"
    variadic:
      min: 2
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+  # String Functions - Case Conversion
+  - source: "string"
+    name: "upper"
+    system_metadata:
+      name: "upper"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
+      - "fixedchar"
```
nielspardon: Impl signatures use [Type Short Names](https://substrait.io/extensions/#function-signature-compound-names), not full type names: `string`→`str`, `varchar`→`vchar`, `fixedchar`→`fchar`, `decimal`→`dec`, `precision_time`→`pt`, `precision_timestamp`→`pts`, `precision_timestamp_tz`→`ptstz`, `interval_day`→`iday`, `interval_year`→`iyear`, `binary`→`vbin`, and a user-defined type → `u!<name>`. `src/custom_extensions.cpp:214-236` already carries this table, and the spec's reference dialect uses `str`/`vchar`.

**88 of the 195 entries** contain at least one long-name impl (160 impl strings in total), so this is the single biggest mechanical fix in the file: **112 of the 193 unresolvable impl strings resolve after nothing but this rename.** `upper` is representative —

```suggestion
    supported_impls:
      - "str"
      - "vchar"
      - "fchar"
```

— and the same rename is needed across the other 35 entries of the `string` block (36 in total), plus `arithmetic_decimal` (1549-1619), `rounding_decimal` (1621-1645), the `datetime` block, and `geometry`.

The other 81 unresolvable strings need more than a rename, and the comments below only reach some of them. To be explicit about what's left over: **24 strings across 19 entries** get no individual comment — `datetime.lt`/`lte`/`gt`/`gte` (`precision_time_precision_time`, no such overload upstream), `datetime.strftime`, `strptime_time`, `strptime_timestamp`, `local_timestamp`, `round_temporal`, `round_calendar`, `logarithmic.log1p` (`i64`), `rounding_decimal.round` (`decimal`, one-arg), `string.repeat`, `replace_slice`, and the five `regexp_*` entries. Same shapes as the cases below — missing `req` tokens, missing required trailing args, wrong widths — but each needs checking against the yaml individually.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+      - "varchar"
+
+  - source: "string"
+    name: "rtrim"
+    system_metadata:
+      name: "rtrim"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string"
+      - "varchar"
```
nielspardon: Beyond the rename, these have no single-argument form upstream — every impl is `str_str` / `vchar_vchar`, the characters-to-trim argument being required in the Substrait declaration even though DuckDB defaults it.

```suggestion
  - source: "string"
    name: "trim"
    system_metadata:
      name: "trim"
      notation: "FUNCTION"
    supported_impls:
      - "str_str"
      - "vchar_vchar"

  - source: "string"
    name: "ltrim"
    system_metadata:
      name: "ltrim"
      notation: "FUNCTION"
    supported_impls:
      - "str_str"
      - "vchar_vchar"

  - source: "string"
    name: "rtrim"
    system_metadata:
      name: "rtrim"
      notation: "FUNCTION"
    supported_impls:
      - "str_str"
      - "vchar_vchar"
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+  - source: "string"
+    name: "lpad"
+    system_metadata:
+      name: "lpad"
+      notation: "FUNCTION"
+    supported_impls:
+      - "string_i32"
+      - "string_i32_string"
+      - "varchar_i32"
+      - "varchar_i32_varchar"
```
nielspardon: Same shape: the two-argument forms (`string_i32`, `varchar_i32`) don't exist upstream — the pad character is part of every declared impl.

```suggestion
    supported_impls:
      - "str_i32_str"
      - "vchar_i32_vchar"
```

The identical change is needed on `rpad` (1152-1156) and `center` (1163-1167). `center` has no DuckDB counterpart either — see C25.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+  - source: "datetime"
+    name: "extract"
+    system_metadata:
+      name: "extract"
+      notation: "FUNCTION"
+    supported_impls:
+      - "date"
+      - "precision_time"
+      - "precision_timestamp"
+      - "precision_timestamp_tz"
```
nielspardon: Enumeration arguments are part of the signature as `req`, so `extract` resolves to `req_date`, `req_pt`, `req_pts`, `req_ptstz_str` — none of the four declared strings exist. (Upstream also has `req_req_*` variants from a second enum arg.) `extract` isn't a DuckDB catalog function either; the DuckDB spelling is `date_part`, since `extract(… FROM …)` is syntax.

```suggestion
  # DateTime Functions - Extract
  - source: "datetime"
    name: "extract"
    system_metadata:
      name: "date_part"
      notation: "FUNCTION"
    supported_impls:
      - "req_date"
      - "req_pt"
      - "req_pts"
      - "req_ptstz_str"
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+
+  - source: "datetime"
+    name: "extract_boolean"
+    system_metadata:
+      name: "extract_boolean"
+      notation: "FUNCTION"
+    supported_impls:
+      - "date"
+      - "precision_timestamp"
```
nielspardon: I'd remove this entry rather than re-sign it. Upstream `extract_boolean`'s only components are `IS_LEAP_YEAR` and `IS_DST`, and it returns `boolean`. DuckDB can't express that: `date_part('is_leap_year', …)` gives `Conversion Error: extract specifier "is_leap_year" not recognized` (same for `is_dst`), `date_part` only ever returns `BIGINT` or `STRUCT`, and there is no function matching `%leap%` in `duckdb_functions()` on 1.5.5. The declared impls (`date`, `precision_timestamp`) are also missing the `req` enum token, so nothing here resolves in either direction.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+    system_metadata:
+      name: "subtract"
+      notation: "FUNCTION"
+    supported_impls:
+      - "date_date"
+      - "precision_timestamp_precision_timestamp"
+      - "precision_timestamp_tz_precision_timestamp_tz"
+      - "date_interval_day"
+      - "precision_timestamp_interval_day"
+      - "precision_timestamp_tz_interval_day"
```
nielspardon: Three of the six `add` impls have the operands reversed (`interval_day_date`); upstream declares only `<temporal>_<interval>`. `subtract` has no `date_date` / `pts_pts` / `ptstz_ptstz` form at all — date difference isn't in `functions_datetime.yaml`, only `subtract(temporal, interval)`. Both are also missing the `_iyear` variants, which `custom_extensions_generated.cpp:18-20` and `:34-37` show the extension registers.

```suggestion
    supported_impls:
      - "date_iday"
      - "date_iyear"
      - "pts_iday"
      - "pts_iyear"
      - "ptstz_iday"
      - "ptstz_iyear_str"

  - source: "datetime"
    name: "subtract"
    system_metadata:
      name: "subtract"
      notation: "FUNCTION"
    supported_impls:
      - "date_iday"
      - "date_iyear"
      - "pts_iday"
      - "pts_iyear"
      - "ptstz_iday"
      - "ptstz_iyear"
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+
+  - source: "datetime"
+    name: "multiply"
+    system_metadata:
+      name: "multiply"
+      notation: "FUNCTION"
+    supported_impls:
+      - "interval_day_i64"
+      - "i64_interval_day"
```
nielspardon: `interval_day_i64` has the operands reversed — upstream `multiply` is only ever `<integer>_<interval>`. Note the rename alone doesn't fix it (`interval_day_i64` → `iday_i64`, which still matches nothing); it needs the operand order corrected too, and once it is, it's identical to line 1275's `i64_interval_day` → `i64_iday`. So the reversed line wants deleting rather than fixing, or you end up with `i64_iday` declared twice.

```suggestion
      - "i64_iday"
```

Upstream also declares the `i8`/`i16`/`i32` widths and the `_iyear` variants (`i8_iday` … `i64_iyear`), which `custom_extensions_generated.cpp:24-31` shows the extension registers — worth adding if DuckDB handles them.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+  rounding_decimal: "extension:io.substrait:functions_rounding_decimal"
+  boolean: "extension:io.substrait:functions_boolean"
+  list: "extension:io.substrait:functions_list"
+  logarithmic: "extension:io.substrait:functions_logarithmic"
+  set: "extension:io.substrait:functions_set"
+
+# ============================================================================
+# SCALAR FUNCTIONS
+# ============================================================================
+supported_scalar_functions:
```
nielspardon, line 206: 158 of the 195 entries have `system_metadata.name` identical to the Substrait `name` — for the field whose whole purpose is the DuckDB mapping. **60 of those 158 don't exist in `duckdb_functions()`** on DuckDB 1.5.5, the submodule pin. (Counting all `notation: FUNCTION` entries regardless of whether the name was copied, it's 63.) Beyond geometry (C26), `least`/`greatest` (C16) and the `is_nan` family (C14):

| line | Substrait | declared DuckDB name | actual DuckDB spelling |
|---|---|---|---|
| 1651 | `index_in` | `index_in` | `list_position` |
| 1668 | `sort` | `sort` | `list_sort` |
| 1676 | `transform` | `transform` | `list_transform` |
| 1692 | `any_match` | `any_match` | — none |
| 1700 | `all_match` | `all_match` | — none |
| — | `filter` (1684) | `filter` | already correct — DuckDB has `filter(ANY[], LAMBDA)` |
| 1740 | `sum0` | `sum0` | — none |
| 1785 | `std_dev` | `std_dev` | `stddev` / `stddev_samp` |

A longer tail — `is_true`/`is_not_true`/`is_false`/`is_not_false`, `capitalize`, `title`, `initcap`, `swapcase`, `center`, `replace_slice`, `logb`, `log1p`, the `regexp_*` family, `strptime_date`/`_time`/`_timestamp`, `assume_timezone`, `local_timestamp`, `round_temporal`, `round_calendar` — has no DuckDB function by the declared name either, but "absent from the catalog" doesn't mean "unsupported", so these need sorting by hand rather than in bulk:

- **DuckDB does support them, under a different spelling or as syntax.** `is_true`/`is_not_true`/`is_false`/`is_not_false` all work as SQL syntax (`SELECT true IS TRUE, false IS NOT TRUE` → `true, true`), so they want a notation change rather than removal. `logb` is DuckDB's two-argument `log(base, x)` (`log(2, 8)` → `3.0`). `assume_timezone` and `local_timestamp` are *both* `timezone(tz, ts)`, selected by argument type: `timezone(tz, TIMESTAMP)` → `TIMESTAMP WITH TIME ZONE` is `assume_timezone`, and `timezone(tz, TIMESTAMPTZ)` → `TIMESTAMP` is `local_timestamp`.
- **Near neighbours worth mapping.** `regexp_match_substring`→`regexp_extract`, `regexp_match_substring_all`→`regexp_extract_all`, `regexp_string_split`→`regexp_split_to_array`, `strptime_*`→`strptime`, `round_temporal`→`date_trunc`.
- **Genuinely no counterpart**, so probably should come out: `sum0`, `any_match`, `all_match`, `capitalize`, `title`, `initcap`, `swapcase`, `center`, `replace_slice`, `log1p`, `regexp_count_substring`, `regexp_strpos`, `round_calendar`, `extract_boolean` (see C22).

Given the volume, generating these sections from the pinned extension YAMLs plus the repo's own mapping tables (`RemapFunctionName`, `TypeShortName`, `custom_extensions_generated.cpp`) would be more robust than hand-authoring, and the check in C5 would keep them honest.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+      - "geometry"
+
+  - source: "geometry"
+    name: "collection_extract"
+    system_metadata:
+      name: "collection_extract"
+      notation: "FUNCTION"
+    supported_impls:
+      - "geometry_i32"
+
```
nielspardon: Are these intended as supported today? I can't find a path that would make them work. To be clear about what is and isn't wrong: all 19 Substrait `name` values are correct — every one is declared in `functions_geometry.yaml`. The problems are on the DuckDB side and in the signatures:

- `duckdb_functions()` on DuckDB 1.5.5 has no `point`, `make_line`, `x_coordinate`, `y_coordinate`, `num_points`, `is_empty`, `is_closed`, `is_simple`, `is_ring`, `is_valid`, `geometry_type`, `dimension`, `envelope`, `centroid`, `buffer`, `flip_coordinates`, `remove_repeated_points`, `minimum_bounding_circle` or `collection_extract` — so `system_metadata.name` resolves for none of the 19.
- The `spatial` extension spells them `ST_X`, `ST_IsRing`, `ST_Centroid`, `ST_Buffer`, … and nothing in `src/` maps those onto the Substrait names.
- 18 of the 19 have unresolvable impls: they're written `geometry` / `geometry_geometry` / `geometry_fp64`, where the short name for a user-defined type is `u!geometry`. (`point` is the exception — its args really are two plain `fp64`, so `fp64_fp64` is already correct.) `collection_extract` is further off than a rename: upstream is `u!geometry` / `u!geometry_i8`, not `geometry_i32`.
- `supported_types` has no `USER_DEFINED` entry for the argument type to stand on. Adding one is not quite free: `#/definitions/type_user_defined` requires `source` and `name` identifying a type declared in a listed dependency, so it would need `functions_geometry.yaml`'s own type declaration.

They do appear in `src/custom_extensions_generated.cpp:119-139`, but that file is generated unconditionally from every impl of every ingested YAML, so it isn't evidence of support either way.

If the answer is "not yet", I'd take the block out and add it back with the `spatial` mapping when that lands — a consumer reading this file would otherwise plan spatial work onto DuckDB and fail at execution. If there is a path I've missed, a pointer to it in a comment here would help, since it isn't visible from `src/`.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+    supported_impls:
+      - "list"
+
+  - source: "list"
+    name: "all_match"
+    system_metadata:
+      name: "all_match"
+      notation: "FUNCTION"
+    supported_impls:
+      - "list"
```
nielspardon: `transform`, `filter`, `any_match` and `all_match` each take a lambda upstream, so the signature is `list_func`, not `list`. (`list.sort` just above at 1665-1671 is already correct — it really is `list`.)

Only `transform` needs its DuckDB name changed: `transform` isn't in the catalog but `list_transform` is. `filter` is already right — DuckDB has a real `filter(ANY[], LAMBDA)` scalar, an exact alias of `list_filter` — so I've left it alone.

```suggestion
  - source: "list"
    name: "transform"
    system_metadata:
      name: "list_transform"
      notation: "FUNCTION"
    supported_impls:
      - "list_func"

  - source: "list"
    name: "filter"
    system_metadata:
      name: "filter"
      notation: "FUNCTION"
    supported_impls:
      - "list_func"

  - source: "list"
    name: "any_match"
    system_metadata:
      name: "any_match"
      notation: "FUNCTION"
    supported_impls:
      - "list_func"

  - source: "list"
    name: "all_match"
    system_metadata:
      name: "all_match"
      notation: "FUNCTION"
    supported_impls:
      - "list_func"
```

`any_match`/`all_match` have no DuckDB equivalent at all — neither name is in the catalog and there's no lambda alias for them — so those two may belong out of the list rather than re-signed.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+supported_aggregate_functions:
+  # Generic Aggregates
+  - source: "aggregate_generic"
+    name: "count"
+    system_metadata:
+      name: "count"
+      notation: "FUNCTION"
+    supported_impls:
+      - "i64"
+      - "any"
```
nielspardon, line 1416: `count` has two impls upstream: zero-arg (`count()`, signature `""`) and `count(any1)` (signature `any`). `i64` is the *return* type.

```suggestion
    supported_impls:
      - ""
      - "any"
```

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+      - "i64"
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "mode"
+    system_metadata:
+      name: "mode"
+      notation: "FUNCTION"
+    supported_impls:
+      - "any"
```
nielspardon: `median` takes a leading `precision` enum (`EXACT`/`APPROXIMATE`), so its signatures are `req_i8`…`req_fp64`; `i64`/`fp64` are return types. `mode` has concrete impls (`i8`…`fp64`), not `any`.

```suggestion
  - source: "arithmetic"
    name: "median"
    system_metadata:
      name: "median"
      notation: "FUNCTION"
    supported_impls:
      - "req_i64"
      - "req_fp64"

  - source: "arithmetic"
    name: "mode"
    system_metadata:
      name: "mode"
      notation: "FUNCTION"
    supported_impls:
      - "i64"
      - "fp64"
```

(`precision` is an enumeration *argument*, already accounted for by the `req` token — not a function-level option, so it doesn't belong in `required_options`.)

`quantile` at 1826-1833 I'd raise as a question rather than suggest a fix. Upstream is `quantile(boundaries enum, precision enum, i64 n, any distribution) -> LIST<any>` — signature `req_req_i64_any`, so `fp32`/`fp64` are certainly wrong. But `n` is a *partition count* and the function returns bin boundaries, whereas DuckDB's `quantile`, `quantile_cont` and `quantile_disc` all take explicit fraction(s) and return values from the input domain. No DuckDB aggregate has that shape, and nothing in `src/` maps one, so the entry may belong out of the list. Note `quantile` itself *is* a real DuckDB function, so the current name is the best of the three — don't swap it for `quantile_cont`.

`string_agg` at 1859 is the same class of error — upstream is `str_str` (value plus separator), not `string`/`varchar`.

---

duckdb_dialect.yaml
```diff
@@ -0,0 +1,1998 @@
...
+    supported_impls:
+      - "fp64"
+
+  - source: "arithmetic"
+    name: "cume_dist"
+    system_metadata:
+      name: "cume_dist"
+      notation: "FUNCTION"
+    supported_impls:
+      - "fp64"
```
nielspardon: `row_number`, `rank`, `dense_rank`, `percent_rank` and `cume_dist` take no arguments, so the signature is the empty string — the spec's reference dialect spells `rank` exactly this way (`supported_impls: [""]`). `i64` / `fp64` are return types.

```suggestion
  - source: "arithmetic"
    name: "row_number"
    system_metadata:
      name: "row_number"
      notation: "FUNCTION"
    supported_impls:
      - ""

  - source: "arithmetic"
    name: "rank"
    system_metadata:
      name: "rank"
      notation: "FUNCTION"
    supported_impls:
      - ""

  - source: "arithmetic"
    name: "dense_rank"
    system_metadata:
      name: "dense_rank"
      notation: "FUNCTION"
    supported_impls:
      - ""

  - source: "arithmetic"
    name: "percent_rank"
    system_metadata:
      name: "percent_rank"
      notation: "FUNCTION"
    supported_impls:
      - ""

  - source: "arithmetic"
    name: "cume_dist"
    system_metadata:
      name: "cume_dist"
      notation: "FUNCTION"
    supported_impls:
      - ""
```

`src/custom_extensions.cpp:241-247` already documents this case and the `name:`-with-empty-suffix convention, including the upstream issue (substrait-io/substrait#1162).

---

README.md
```diff
@@ -2,6 +2,7 @@
 
 The DuckDB Substrait Extension is a [DuckDB Community Extension](https://duckdb.org/community_extensions/) that provides [Substrait](https://substrait.io) support to [DuckDB](https://www.duckdb.org).
 With this extension, DuckDB can produce Substrait plans from DuckDB queries as well as consume Substrait plans and execute them with DuckDB.
+The most commonly used logical relations and functions are supported, including those generated from all of the TPC-H and TPC-DS queries. The supported features are formally specified in the Substrait dialect file [duckdb_dialect.yaml](duckdb_dialect.yaml).
```
nielspardon: "including those generated from all of the TPC-H and TPC-DS queries" is accurate for the producer, but both `test/sql/test_substrait_tpch.test` and `test_substrait_tpcds.test` run their queries as `statement ok` on `CALL get_substrait(...)` only — no `from_substrait`, no result comparison. Since the sentence sits next to a claim about *supported features*, it's worth being precise about which direction is covered.

```suggestion
The most commonly used logical relations and functions are supported: Substrait plans can be produced for all of the TPC-H and TPC-DS queries. The supported features are formally specified in the Substrait dialect file [duckdb_dialect.yaml](duckdb_dialect.yaml).
```

nielspardon (changes requested):
Thanks for this — a machine-readable capability declaration is exactly what consumers need, and wiring it into CI as a validated artifact rather than prose is the right instinct. The structure of the file is good and it conforms to `dialect_schema.yaml` cleanly.

The substance needs work before it can be trusted, though, and one build-side change is load-bearing. Details are inline; the two themes:

**1. The schema is never found where the build looks for it.** `substrait-protobuf` doesn't install `dialect_schema.yaml` — at the pinned 0.98.0 it installs `include/substrait/*.pb.h`, the `.proto` tree into `share/substrait/proto/`, and the CMake config. So the primary lookup path can't hit, and the fallback reaches into `$VCPKG_ROOT/buildtrees/`, which needs `VCPKG_ROOT` set and doesn't exist at all on a binary-cache hit. With `FATAL_ERROR` that aborts configuration of the whole extension. The `substrait-extensions` port (already whitelisted in `vcpkg-configuration.json`) ships both `text/dialect_schema.yaml` and `extensions/*.yaml` and exposes `SubstraitExtensions_DATA_DIR` — I verified the install layout locally. See C2/C3.

**2. About half the declared function implementations don't identify a real implementation.** 193 of 388 `supported_impls` strings don't resolve against `cpp/substrait-extensions/v0.98.0` — the tag `scripts/generate_custom_functions.py:16` already pins. Roughly:

- 112 are the short-name convention: signatures use `str`/`vchar`/`dec`/`pts`/`iday`/`u!geometry`, not the full type names. `src/custom_extensions.cpp:214-236` already carries this table.
- The rest need real corrections — missing `req` tokens for enumeration args, variadic args counted twice, return types used in place of argument signatures, reversed operands, overloads that don't exist upstream.

Relatedly, `system_metadata.name` is identical to the Substrait name in 158 of 195 entries, and 60 of those name DuckDB functions that aren't in `duckdb_functions()` on 1.5.5. The `least`/`greatest` pair is the sharpest case — it inverts the mapping that 733af82 landed on this branch a few commits ago, with `src/to_substrait.cpp:51-56` and `src/from_substrait.cpp:48-69` both documenting the correct direction.

There are also a handful of capability claims that don't match `src/`: `SWITCH`, `WINDOW_FUNCTION`-as-expression and `SUBQUERY` are declared but unsupported; `SINGULAR_OR_LIST`, `NESTED`, `CONSISTENT_PARTITION_WINDOW`, `LEFT_SINGLE`, the `VIRTUAL_TABLE`/`LOCAL_FILES` read types and the three interval types are supported but omitted — and since these lists are closed, an omission is itself a claim.

Applying the inline suggestions takes the impl count from 193/388 unresolvable to 0/360.

Two suggestions worth a second opinion rather than a straight apply: whether the geometry block belongs in the file at all (C26 — DuckDB core has none of the 19 names and `spatial` spells them `ST_*`), and what to do with `quantile`, whose Substrait shape has no DuckDB counterpart (C29).

On method, so the numbers are checkable: signatures were resolved against the pinned extension YAMLs (short names per the spec's Type Short Names table, `req` for enumeration args, variadic counted once, `""` for zero-arg); DuckDB names against `duckdb_functions()` on the pinned submodule; and every suggestion was spliced into the file and re-validated against `dialect_schema.yaml`. Happy to share the scripts if useful — a version of the first two would make a good addition to the new test, since schema conformance alone can't catch any of this (C5).

test/c/test_dialect_validation.cpp
```diff
@@ -0,0 +1,26 @@
...
+	json schema = tojson::loadyaml(SCHEMA_PATH);
+
+	json_validator validator;
+	validator.set_root_schema(schema);
+	try {
+		validator.validate(dialect);
+	} catch (const std::exception &e) {
+		FAIL(DIALECT_PATH << " does not conform to " << SCHEMA_PATH << ": " << e.what());
+	}
+}
```
nielspardon, line 26: Follow-up rather than a change request — the file is clean now, and I don't think landing a data file should be gated on someone building a verification harness first. I have the implementation and I'm happy to open it as a separate PR; flagging it here so it's on record.

Schema conformance can't see that a `supported_impls` string names no real implementation. All 193 unresolvable signatures at `bd86a69` passed this test, and it would pass just as happily if they came back. The pin moves with `SUBSTRAIT_EXTENSIONS_TAG`, so each bump can silently reintroduce the drift.

Two ways to close it, both using data already in place:

**Against the extension YAMLs** — complete, and needs no new CMake. `SUBSTRAIT_DATA_DIR` already points at the root holding `extensions/*.yaml` alongside `text/dialect_schema.yaml`, and `tojson::loadyaml` already loads from there. For each `(source, name, impl)`: resolve `dependencies[source]` to the extension file by `urn`, find `name` under the matching function kind, and require `impl` to equal one impl's argument signature — short names per [the compound-name rules](https://substrait.io/extensions/#function-signature-compound-names), `req` per enumeration argument, the variadic argument counted once, `""` for a zero-argument impl. That last set of rules is the whole trick; getting `req` and the variadic count wrong is what makes hand-checking these unreliable.

**Against `custom_extensions_generated.cpp`** — weaker but needs nothing external. Require every declared impl to appear as an `Insert{,Variadic}CustomFunction` registration under the matching URN. I measured this one: it catches 189 of the 193. The four it misses are enumeration-argument cases — `median` declared as `fp64` where upstream is `req_fp64`, `extract` as `date` where upstream is `req_date` — because the registry key omits enum arguments, so `fp64` and `req_fp64` collapse to the same key. It has the nice property of being a pure in-repo invariant: no network, no data dir, no pinned tag.

The first is the better check and the data's already wired up, so that's the one I'd write. Since it needs the same YAML parsing `scripts/generate_custom_functions.py` already does, `scripts/` next to the generator may be the lower-friction home for it rather than C++ in `test_substrait_exe` — happy either way.

nielspardon (approved):
Approving. Everything I raised is addressed, and the dialect now holds up under every check I can construct against it.

Re-verified at `795dfd1` against `substrait-extensions` `v0.98.0` — the same pin `SUBSTRAIT_EXTENSIONS_TAG` uses, so this is checking the file against the definitions the generated registry was built from:

| check | `bd86a69` | `795dfd1` |
| --- | --- | --- |
| schema errors (Draft 2020-12) | 0 | 0 |
| `supported_impls` that resolve to no upstream implementation | 193 / 388 | **0 / 322** |
| `supported_impls` absent from `custom_extensions_generated.cpp` | 189 / 388 | **0 / 322** |
| impl type short names with no `supported_types` entry | 4 | **0** |
| `system_metadata.name` (notation `FUNCTION`) absent from `duckdb_functions()` | 4 | **2** |
| orphaned `dependencies` entries | 0 | 0 |

The two remaining catalog absences are `BETWEEN` and `coalesce`, both correct — neither appears in `duckdb_functions()` because they're grammar rather than catalog entries.

Two things I checked more carefully this round, since they're where I made mistakes the first time:

**Relations and expressions now match the consumer exactly.** `supported_relations` is precisely the 13 cases `TransformOp` accepts, and `supported_expressions` precisely the 7 `TransformExpr` accepts — with `SUBQUERY` correctly gone, since its case label falls through to that switch's throwing `default`. My earlier claim that seven relations were "consumed but omitted" was wrong: I'd read `GetCommon`, which enumerates rel types only to return the `RelCommon` submessage for emit mapping. Sorry for the noise there.

**A concern I had in reserve turns out not to be one.** A handful of declared signatures carry a trailing argument DuckDB's function has no parameter for — `strftime`'s `ptstz_str_str` and `extract`'s `req_ptstz_str` both end in `timezone`, and `strptime_timestamp`'s `str_str_str_i8` ends in `precision`. I was going to raise it, then found all of them registered verbatim in `custom_extensions_generated.cpp`. That file is generated wholesale from the upstream YAMLs, so this is a property of the existing registry, not something the dialect introduces. Worth a separate look at some point; nothing to change here.

One follow-up, non-blocking — inline on the test file. There's still no gate that would have caught any of the 193, and I'd rather write it than ask you to.

PR #265, opened 2026-08-24
build: update Substrait to v0.102.0

### LINKED ISSUE #264: Update to Substrait v0.102.0
# Release Notes

### ⚠ BREAKING CHANGES

* **protos:** deprecates the `offset` field of the `Preceding` and
`Following` window bounds in favor of the new `offset_expr` field, which
accepts an arbitrary expression instead of an integer literal. Consumers
should add `offset_expr` support before producers switch to it. The PR
also requires explicit window frame bounds type. Producers should set
the bounds type and Consumers should reject `BOUNDS_TYPE_UNSPECIFIED` in
such cases.
 

### Features

* **extensions:** add a decimal overload for negate ([#1177](https://github.com/substrait-io/substrait/issues/1177)) ([6092523](https://github.com/substrait-io/substrait/commit/60925234a0a2c02a41400df349b0bad1e674581b))
* **protos:** support expression in window aggregate bounds ([#1105](https://github.com/substrait-io/substrait/issues/1105)) ([6706015](https://github.com/substrait-io/substrait/commit/6706015ffdb19345fef399ef8238e3e6269cb380)), closes [#748](https://github.com/substrait-io/substrait/issues/748) [#617](https://github.com/substrait-io/substrait/issues/617) [#930](https://github.com/substrait-io/substrait/issues/930)

See https://github.com/substrait-io/substrait/releases/tag/v0.102.0

bvolpato:
## Summary

Update extension dependencies, generated registrations, and window-bound handling to Substrait v0.102.0.

## Changes

- Pin vcpkg registry and Substrait protobuf/extensions packages to 0.102.0.
- Regenerate custom function registrations, including decimal negate.
- Advertise produced plans as Substrait 0.102.0.
- Emit `offset_expr` plus compatible legacy `offset` values during migration.
- Prefer `offset_expr` when reading bounds, validate legacy offsets, and require bounds type only for current-row or offset bounds.
- Normalize zero offsets to `CurrentRow` and reject negative offsets before serialization.
- Cover bounded, following, legacy, missing-offset, unbounded, zero, and negative window frames.

## Testing

- `cmake --build build/release --config Release --parallel 8`
- `build/release/test/unittest 'test/sql/*'` (53 test cases, 1,596 assertions)
- `./build/release/extension/substrait/test/c/test_substrait_exe` (50 test cases, 261 assertions)

Closes #264


src/from_substrait.cpp
```diff
@@ -1635,8 +1635,33 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformWindowOp(const substrait::Rel &
 		
 		// Handle window bounds
 		if (window_func.has_lower_bound() || window_func.has_upper_bound()) {
```
nielspardon, line 1637: Set both boundaries to the partition edges before this block.

`WindowExpression::start`/`end` default to `WindowBoundary::INVALID`, so a `Bound` whose `kind` oneof is unset — or a `WindowFunction` carrying neither bound — falls through every branch below and reaches `throw InternalException("Unsupported window start boundary")` in `window_boundaries_state.cpp`, which invalidates the database for the rest of the connection. Defaulting to unbounded is also what the field docs ask for: an absent bound defaults to the start or end of the partition.

```suggestion
		// Handle window bounds
		// An absent bound means the start or end of the partition; keep the boundaries out of
		// WindowBoundary::INVALID so a Bound with no kind set cannot reach the executor.
		window_expr->start = WindowBoundary::UNBOUNDED_PRECEDING;
		window_expr->end = WindowBoundary::UNBOUNDED_FOLLOWING;
		if (window_func.has_lower_bound() || window_func.has_upper_bound()) {
```

src/from_substrait.cpp
```diff
@@ -1635,8 +1635,33 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformWindowOp(const substrait::Rel &
...
+			auto transform_offset = [this](const auto &bound) -> unique_ptr<ParsedExpression> {
+				if (bound.has_offset_expr()) {
+					return TransformExpr(bound.offset_expr());
+				}
+				auto offset = bound.offset();
+				if (offset <= 0) {
+					throw InvalidInputException("Window bound must specify a positive offset or offset expression");
+				}
+				return make_uniq<ConstantExpression>(Value::BIGINT(offset));
+			};
```
nielspardon, line 1664: Route `offset_expr` through the existing `ExtractLiteralInteger` helper instead of `TransformExpr`.

An `offsetExpr` with no `rex_type` set lands in `TransformExpr`'s default branch, where `FindFieldByNumber(sexpr.rex_type_case())` is called with `0`, returns null, and `->name()` dereferences it — the error path itself segfaults. The helper also gives you the NULL and negative rejections the field requires, and keeping offsets literal-only matches both the `FetchRel` path and what this producer emits.

```suggestion
			auto transform_offset = [](const auto &bound) -> unique_ptr<ParsedExpression> {
				if (bound.has_offset_expr()) {
					auto offset = ExtractLiteralInteger(bound.offset_expr(), "window bound offset");
					return make_uniq<ConstantExpression>(Value::BIGINT(static_cast<int64_t>(offset)));
				}
				auto offset = bound.offset();
				if (offset <= 0) {
					throw InvalidInputException("Window bound must specify a positive offset or offset expression");
				}
				return make_uniq<ConstantExpression>(Value::BIGINT(offset));
			};
```

src/to_substrait.cpp
```diff
@@ -1757,8 +1757,19 @@ substrait::Rel *DuckDBToSubstrait::TransformWindow(LogicalOperator &dop) {
 					// TODO: Support expression-based offsets
 					if (boundary_expr->GetExpressionClass() == ExpressionClass::BOUND_CONSTANT) {
 						auto &const_expr = boundary_expr->Cast<BoundConstantExpression>();
-						auto preceding = bound->mutable_preceding();
-						preceding->set_offset(const_expr.value.GetValue<int64_t>());
+						auto offset = const_expr.value.GetValue<int64_t>();
```
nielspardon, line 1760: Guard the NULL case before `GetValue<int64_t>()`.

`ROWS BETWEEN NULL PRECEDING AND CURRENT ROW` binds to a NULL BIGINT constant, and `Value::GetValue<int64_t>()` on a NULL value throws `InternalException`, which invalidates the database instead of reporting a plan error. The FOLLOWING copy just below needs the same guard.

```suggestion
						auto &const_expr = boundary_expr->Cast<BoundConstantExpression>();
						if (const_expr.value.IsNull()) {
							throw NotImplementedException("NULL window bound offsets are not supported");
						}
						auto offset = const_expr.value.GetValue<int64_t>();
```

test/sql/test_window.test
```diff
@@ -225,3 +225,105 @@ SELECT * FROM from_substrait_json(getvariable('plan_cume_dist'))
...
+9	110
+10	110
+
+statement ok
+SET VARIABLE zero_offset_plan = (SELECT json FROM get_substrait_json('SELECT id, sum(val) OVER (PARTITION BY part1 ORDER BY id ROWS BETWEEN 0 PRECEDING AND CURRENT ROW) AS total FROM window_test_data'))
+
+query TT
+SELECT contains(getvariable('zero_offset_plan'), '"currentRow"')::VARCHAR, contains(getvariable('zero_offset_plan'), '"i64":"0"')::VARCHAR
+----
+true	false
```
nielspardon, line 324: Assert that `"preceding"` is absent instead.

Both columns pass no matter what the producer does: `"currentRow"` is already in the plan for the upper bound, and a zero `offset` is proto3's default so `"i64":"0"` would not appear in the JSON even if the zero mapping regressed. Checking that no `preceding` bound was emitted is what actually pins the CurrentRow rewrite.

```suggestion
query TT
SELECT contains(getvariable('zero_offset_plan'), '"preceding"')::VARCHAR, contains(getvariable('zero_offset_plan'), '"offsetExpr"')::VARCHAR
----
false	false
```

nielspardon (changes requested):
The first three suggestions are merge blockers: an empty `offsetExpr`, a NULL ROWS offset, and a `Bound` with no kind set each crash or invalidate the database instead of raising a plan error. GROUPS frames are relabelled ROWS outside this diff and now also receive an `offset_expr`, so a round trip silently returns different rows — I'll file that separately. Test-coverage and version-bump notes will follow as a second round.

nielspardon:
@bvolpato I know you've already signed the CLA for other repos. Looks like the CLA assistant hasn't carried that over and you need to sign once more for this repo.

test/sql/test_window.test
```diff
@@ -225,3 +225,128 @@ SELECT * FROM from_substrait_json(getvariable('plan_cume_dist'))
...
+7	95
+8	95
+9	110
+10	110
+
+statement ok
+CREATE TABLE window_timestamp_data (ts TIMESTAMP, val INTEGER)
+
+statement ok
+INSERT INTO window_timestamp_data VALUES ('2024-01-01 00:00:00', 1), ('2024-01-02 00:00:00', 2)
```
EpsilonPrime: With only two rows one day apart, a 1-day RANGE frame and an unbounded frame give the same totals (1, 3), so this test would pass even if the interval offset were ignored. Could you add a third row further out, e.g. `('2024-01-04 00:00:00', 4)`? Its total would be 4 with the 1-day frame and 7 if the offset were dropped.

src/to_substrait.cpp
```diff
@@ -1797,8 +1797,22 @@ substrait::Rel *DuckDBToSubstrait::TransformWindow(LogicalOperator &dop) {
 					// TODO: Support expression-based offsets
 					if (boundary_expr->GetExpressionClass() == ExpressionClass::BOUND_CONSTANT) {
 						auto &const_expr = boundary_expr->Cast<BoundConstantExpression>();
-						auto preceding = bound->mutable_preceding();
-						preceding->set_offset(const_expr.value.GetValue<int64_t>());
+						if (const_expr.value.IsNull()) {
+							throw NotImplementedException("NULL window bound offsets are not supported");
+						}
+						auto offset = const_expr.value.GetValue<int64_t>();
```
EpsilonPrime: Now that the consumer keeps typed RANGE offsets, the producer is the lossy side. For a RANGE frame with a fractional offset (`RANGE BETWEEN 1.5 PRECEDING ...`), `GetValue<int64_t>()` rounds it, producing a plan with a different frame. An INTERVAL offset throws a conversion error. Could the RANGE case emit the constant as a typed literal in `offset_expr`, setting the legacy `offset` only for integer values? Otherwise, reject non-integer RANGE offsets explicitly. This predates the PR, so a follow-up is fine if you'd rather not grow it. (The FOLLOWING branch below has the same pattern.)

src/to_substrait.cpp
```diff
@@ -1797,8 +1797,22 @@ substrait::Rel *DuckDBToSubstrait::TransformWindow(LogicalOperator &dop) {
 					// TODO: Support expression-based offsets
```
EpsilonPrime: Nit: the `// For now, we only support constant integer offsets` / `// TODO: Support expression-based offsets` comments are now partly stale, since `offset_expr` is emitted. Worth updating (the same comment appears again at the FOLLOWING branch).

PR #247, opened 2026-07-23
feat: implement support for missing Substrait literal types

andrew-coleman:
Continued from PR #240, which got corrupted for some reason.
See #240 for discussion

PR #269, opened 2026-09-03
fix: empty-table optimizer folding in `get_substrait`

### LINKED ISSUE #201: Optimizer folds empty tables into non-portable virtualTable plans
### What happens?

When `get_substrait_json` is called with the optimizer enabled (default) against a table that has 0 rows, the optimizer produces a `LogicalEmptyResult` which serializes as a `virtualTable {}` with no `namedTable` reference. The resulting Substrait plan is **non-portable** — it always returns 0 rows regardless of the actual data at execution time.

In `substrait_extension.cpp`, `ExtractPlan` runs the full DuckDB optimizer (line 84):

```cpp
if (context.config.enable_optimizer) {
    Optimizer optimizer(*planner.binder, context);
    plan = optimizer.Optimize(std::move(plan));
}
```

The optimizer's statistics propagation pass sees 0 rows in the source table and folds the entire logical plan into `LogicalEmptyResult`. Then `TransformEmptyResult` in `to_substrait.cpp` (line 1974) serializes this as:

```json
{"read": {"baseSchema": {...}, "virtualTable": {}}}
```

This plan has no `namedTable` — the table reference is gone. Executing it will always return 0 rows.

## Why This Matters

Substrait plans are designed to be portable — compiled against a schema and executed against any data conforming to that schema. Use cases include:

- Compiling against schema-only tables (no data available at compile time) then executing remotely
- Caching compiled plans and reusing them as data changes
- Cross-engine execution (compile on DuckDB, execute on another engine)

The optimizer's empty-table folding breaks this portability contract.

## Workaround

Pass `enable_optimizer := false`:

```sql
CALL get_substrait_json('SELECT ...', enable_optimizer := false)
```

This produces a correct, portable plan but sacrifices all optimization.

### To Reproduce

```cpp
DuckDB db(nullptr);
Connection con(db);

// Create table (empty)
con.Query("CREATE TABLE alerts (id INTEGER, status VARCHAR, reason VARCHAR)");

// Compile plan while empty — optimizer folds to virtualTable {}
auto plan_json = GetSubstraitJSON(con, 
    "SELECT id, reason FROM alerts WHERE status = 'PENDING' ORDER BY id");

// Insert data
con.Query("INSERT INTO alerts VALUES (1, 'PENDING', 'Large transfer'), "
          "(2, 'RESOLVED', 'False alarm'), (3, 'PENDING', 'Suspicious pattern')");

// Execute the plan — returns 0 rows instead of 2
auto result = FromSubstraitJSON(con, plan_json);
// Expected: rows (1, 'Large transfer'), (3, 'Suspicious pattern')
// Actual: 0 rows
```

### OS:

OSX

latest

### DuckDB Version:

1.4.3

### DuckDB Client:

C++/DuckDB/Python

- [x] I have

- [x] I have

thgschwind:
When `get_substrait` / `get_substrait_json` compiles a query while the referenced table(s) are **empty at compile time**, DuckDB's optimizer uses the zero cardinality to fold the whole query into a `LogicalEmptyResult`. This serializes as a Substrait `read` on a `virtualTable {}` (an inline table with no rows) instead of a `namedTable` scan, so the plan always returns 0 rows, regardless of the data present when the plan is later executed.

This breaks the portability contract of a Substrait plan: it should be valid for *any* data conforming to the schema, not just the table state at compile time. This bites any workflow that pre-compiles in a different environment, a primary reason to export Substrait in the first place.

### Fix

Disable the two optimizer passes responsible for the folding in `ToSubstraitFunctionData` (`src/substrait_extension.cpp`), alongside the optimizers already disabled there (`IN_CLAUSE`, `COMPRESSED_MATERIALIZATION`, `COMMON_SUBPLAN`, …):

```cpp
disabled_optimizers.insert(OptimizerType::STATISTICS_PROPAGATION);
disabled_optimizers.insert(OptimizerType::EMPTY_RESULT_PULLUP);
```

- **`STATISTICS_PROPAGATION`** is the pass that actually triggers the bug: it reads table cardinality during optimization and, on an empty table, derives a zero-row estimate and folds the query into `LogicalEmptyResult`.
- **`EMPTY_RESULT_PULLUP`** is disabled as a safeguard: it propagates a `LogicalEmptyResult` upward through joins/projections, so if any other pass introduces an empty-result node in the future, it won't cascade into a fully-folded plan.

The implementation is more narrow than the suggestion in #201 (`enable_optimizer := false`).  While this produces a correct plan, it would also throw away *all* other optimizations. This change instead disables only the two passes tied to this issue, so every other optimization (filter pushdown, projection pruning, join ordering, etc.) still runs and the emitted plan stays both portable and optimized.

### Testing

Adds a C/Catch regression test, `test/c/test_empty_table_optimizer_folding.cpp`, with two cases:

1. **Single table**: compile `SELECT id, reason FROM alerts WHERE status = 'PENDING' ORDER BY id` against an empty table, insert three rows, execute the compiled plan, and assert the two matching rows come back.
2. **Join** (the more common real-world case): same pattern over a `customers ⋈ orders` join with a filter, asserting the expected rows after data is inserted.

Both tests fail on `main` (the plan folds to empty and returns 0 rows) and pass with this change. Verified against current `main` (tip `4ea63dd`).

Wired into the C test suite via `test/c/CMakeLists.txt`.

Fixes #201

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Substrait plans compiled against empty tables so they continue returning correct results after data is inserted.
  * Corrected behavior for both filtered queries and joins involving tables that were initially empty.

* **Tests**
  * Added regression coverage for empty-table query planning.
  * Updated UUID round-trip testing to account for currently unsupported UUID literal handling.


src/substrait_extension.cpp
```diff
@@ -56,13 +56,25 @@ struct ToSubstraitFunctionData : public TableFunctionData {
...
+		// it won't cascade into a fully-folded plan.
 		set<OptimizerType> disabled_optimizers = DBConfig::GetConfig(context).options.disabled_optimizers;
 		disabled_optimizers.insert(OptimizerType::IN_CLAUSE);
 		disabled_optimizers.insert(OptimizerType::COMPRESSED_MATERIALIZATION);
 		disabled_optimizers.insert(OptimizerType::MATERIALIZED_CTE);
 		// The CommonSubplan optimizer deduplicates repeated subplans into
 		// materialized CTEs, which cannot be expressed in Substrait.
 		disabled_optimizers.insert(OptimizerType::COMMON_SUBPLAN);
+		disabled_optimizers.insert(OptimizerType::STATISTICS_PROPAGATION);
+		disabled_optimizers.insert(OptimizerType::EMPTY_RESULT_PULLUP);
```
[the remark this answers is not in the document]
thgschwind: The UUID gap is pre-existing, not created by this PR. A selective UUID filter already throws on today's main; STATISTICS_PROPAGATION only masks it for non-selective filters by folding the comparison away before it reaches the producer. Disabling stats-prop widens exposure of an existing limitation, it doesn't introduce a new bug. The consumer intentionally maps UUID literals to BLOB (PR #247) because a virtualTable read of a typed UUID column would need an implicit BLOB to UUID cast DuckDB doesn't provide, so faithful round-trip is a design change, not a missing-case fix.

src/substrait_extension.cpp
```diff
@@ -56,13 +56,25 @@ struct ToSubstraitFunctionData : public TableFunctionData {
...
+		//
+		// STATISTICS_PROPAGATION must be disabled because it uses table
+		// cardinality at plan-generation time to fold queries against empty
+		// tables into LogicalEmptyResult — producing a non-portable plan
+		// (virtualTable {}) that returns 0 rows regardless of runtime data.
+		// A Substrait plan must be valid for any data conforming to the schema.
+		// EMPTY_RESULT_PULLUP is disabled as an additional safeguard: it
+		// propagates LogicalEmptyResult upward through joins/projections, so
+		// if any other pass introduces an empty-result node in the future,
+		// it won't cascade into a fully-folded plan.
```
EpsilonPrime: This explanation sits above the `IN_CLAUSE` insert, several lines from the two inserts it describes, while the COMMON_SUBPLAN comment sits directly above its own insert. Could you move it down next to the `STATISTICS_PROPAGATION` and `EMPTY_RESULT_PULLUP` inserts?

test/c/test_empty_table_optimizer_folding.cpp
```diff
@@ -0,0 +1,64 @@
+#include "catch.hpp"
+#include "test_helpers.hpp"
+#include "duckdb/main/connection_manager.hpp"
```
EpsilonPrime: Nit: `duckdb/main/connection_manager.hpp` appears unused.

EpsilonPrime (commented):
This also fixes statistics-based folding on non-empty tables (e.g. `WHERE id > 100` when the max id is 3). Consider adding a case for that (insert 1..3, compile, insert 101, execute), and mentioning it in the description.

PR #271, opened 2026-09-03
fix: map substrait extract(UNIX_TIME) to date_part('epoch')

thgschwind:
The Substrait standard defines `extract` with a `component` enum whose only numeric date/timestamp difference is **`UNIX_TIME`** (the number of seconds elapsed since `1970-01-01 00:00:00 UTC`) returned as `i64` (see `functions_datetime.yaml`).

The DuckDB extension does not support `extract(UNIX_TIME)`. In `SubstraitToDuckDB` (`src/from_substrait.cpp`), because the `extract` branch runs every specifier through `VerifyCorrectExtractSubfield`, which checks membership in `valid_extract_subfields`, a set that does not include `UNIX_TIME`, and DuckDB has no `date_part` specifier of that name. So:

- In debug builds, `VerifyCorrectExtractSubfield`'s `D_ASSERT(...count(subfield))` fires.
- In release builds the assert is compiled out, so the transform builds `date_part('UNIX_TIME', <expr>)`, which DuckDB rejects at binding.

### Fix

Special-case `UNIX_TIME` in the `extract` branch of `SubstraitToDuckDB::TransformScalarFunctionExpr` (`src/from_substrait.cpp`): map it to DuckDB's `date_part('epoch', <expr>)`, which yields the epoch seconds as `DOUBLE`, and wrap the result in a cast to `BIGINT` to honor the standard's `i64` return type.

```cpp
if (subfield == "UNIX_TIME") {
    children.insert(children.begin(), make_uniq<ConstantExpression>(Value("epoch")));
    auto call = make_uniq<FunctionExpression>("date_part", std::move(children));
    return make_uniq<CastExpression>(LogicalType::BIGINT, std::move(call));
}
```

All other `extract` specifiers are unchanged — they continue through `VerifyCorrectExtractSubfield` and the normal `date_part` path. Only the previously-unhandled `UNIX_TIME` specifier is affected.

### Testing

Full build against DuckDB v1.5.5, with both test suites green and no regressions:

- **C unit tests** (`test_substrait_exe`): 261 assertions / 50 cases pass.
- **SQL suite** (`test/*`): 1501 assertions / 53 cases pass, including `test_substrait_extract.test` which exercises the `extract` path.

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -607,12 +607,24 @@ unique_ptr<ParsedExpression> SubstraitToDuckDB::TransformScalarFunctionExpr(cons
 		D_ASSERT(enum_expressions.size() == 1);
 		auto &subfield = enum_expressions[0];
 		// Substrait's UNIX_TIME specifier (epoch seconds, return type i64) has no
-		// DuckDB date_part equivalent of that name. Map it to date_part('epoch'),
-		// which yields DOUBLE, and cast back to BIGINT to honor the standard's i64.
-		if (subfield == "UNIX_TIME") {
+		// DuckDB date_part equivalent of that name. Map it to date_part('epoch') and
+		// floor before casting, to honor the standard's "elapsed whole seconds".
+		if (StringUtil::CIEquals(subfield, "UNIX_TIME")) {
+			// The precision_timestamp_tz impl carries a trailing timezone argument.
+			// UNIX_TIME is timezone-independent, so drop it instead of emitting an
+			// unbindable three-argument date_part.
+			if (children.size() > 1) {
+				children.resize(1);
+			}
 			children.insert(children.begin(), make_uniq<ConstantExpression>(Value("epoch")));
-			auto call = make_uniq<FunctionExpression>("date_part", std::move(children));
-			return make_uniq<CastExpression>(LogicalType::BIGINT, std::move(call));
+			auto call = make_uniq<FunctionExpression>(RemapFunctionName(function_name), std::move(children));
+			// date_part('epoch') returns DOUBLE and DuckDB's DOUBLE->BIGINT cast rounds
+			// half-to-even, which would report one second too many for any sub-second
+			// timestamp. UNIX_TIME is elapsed whole seconds, so floor explicitly.
+			vector<unique_ptr<ParsedExpression>> floor_args;
+			floor_args.push_back(std::move(call));
+			auto floored = make_uniq<FunctionExpression>("floor", std::move(floor_args));
+			return make_uniq<CastExpression>(LogicalType::BIGINT, std::move(floored));
 		}
 		VerifyCorrectExtractSubfield(subfield);
 		auto constant_expression = make_uniq<ConstantExpression>(Value(subfield));
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -606,6 +606,14 @@ unique_ptr<ParsedExpression> SubstraitToDuckDB::TransformScalarFunctionExpr(cons
 	} else if (function_name == "extract") {
 		D_ASSERT(enum_expressions.size() == 1);
 		auto &subfield = enum_expressions[0];
+		// Substrait's UNIX_TIME specifier (epoch seconds, return type i64) has no
+		// DuckDB date_part equivalent of that name. Map it to date_part('epoch'),
+		// which yields DOUBLE, and cast back to BIGINT to honor the standard's i64.
+		if (subfield == "UNIX_TIME") {
+			children.insert(children.begin(), make_uniq<ConstantExpression>(Value("epoch")));
+			auto call = make_uniq<FunctionExpression>("date_part", std::move(children));
+			return make_uniq<CastExpression>(LogicalType::BIGINT, std::move(call));
+		}
 		VerifyCorrectExtractSubfield(subfield);
 		auto constant_expression = make_uniq<ConstantExpression>(Value(subfield));
 		children.insert(children.begin(), std::move(constant_expression));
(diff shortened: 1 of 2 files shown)


src/from_substrait.cpp
```diff
@@ -606,6 +606,14 @@ unique_ptr<ParsedExpression> SubstraitToDuckDB::TransformScalarFunctionExpr(cons
...
 		D_ASSERT(enum_expressions.size() == 1);
 		auto &subfield = enum_expressions[0];
+		// Substrait's UNIX_TIME specifier (epoch seconds, return type i64) has no
+		// DuckDB date_part equivalent of that name. Map it to date_part('epoch'),
+		// which yields DOUBLE, and cast back to BIGINT to honor the standard's i64.
+		if (subfield == "UNIX_TIME") {
+			children.insert(children.begin(), make_uniq<ConstantExpression>(Value("epoch")));
+			auto call = make_uniq<FunctionExpression>("date_part", std::move(children));
+			return make_uniq<CastExpression>(LogicalType::BIGINT, std::move(call));
+		}
```
andrew-coleman, line 628: Confirms the premise — `extract(UNIX_TIME)` does fail today (`Conversion Error: extract specifier "UNIX_TIME" not recognized` in release, `D_ASSERT` in debug) and mapping it to `date_part('epoch')` is the right call. Four things about the new branch itself, all inside these lines:

**1. The cast rounds; `UNIX_TIME` is elapsed *whole* seconds.** `date_part('epoch')` returns `DOUBLE`, and DuckDB's `DOUBLE`→`BIGINT` cast uses `std::nearbyint`, so any sub-second timestamp reports one second too many:

| timestamp | current | spec |
|---|---|---|
| `1970-01-01 00:00:00.75` | `1` | `0` |
| `1970-01-01 00:00:01.5` | `2` | `1` |
| `1969-12-31 23:59:59.5` | `0` | `-1` |
| `2021-08-03 11:59:44.75` | `1627991985` | `1627991984` |

Note `.5 → 0` but `1.5 → 2` — half-to-even, so it isn't even consistently off. Needs an explicit `floor()`.

**2. The comparison is case-sensitive** while every other subfield check here isn't — `valid_extract_subfields` is a `case_insensitive_set_t`, and this extension's own producer emits *lowercase* enum values (`to_substrait.cpp` sets the enum to the DuckDB function name). A plan carrying `unix_time` skips the branch and lands back on the broken path.

**3. `RemapFunctionName` already owns the `"date_part"` name.** `function_names_remap` has `{"extract", "date_part"}` and the fall-through at the end of the function uses it; the literal here means the mapping now lives in two places.

**4. `UNIX_TIME` still won't work for one of the impls you're fixing.** Substrait's `precision_timestamp_tz` signature is `extract(component, x, timezone: string)` — `timezone` is a *value* arg, so it lands in `children` and you get `date_part('epoch', x, tz)` → `Binder Error: No function matches date_part(VARCHAR, TIMESTAMPTZ, VARCHAR)`. This repo's own `duckdb_dialect.yaml:1124` advertises `req_ptstz_str` as supported. To be clear this one is **pre-existing** — the same code shape breaks `YEAR`/`MONTH`/`HOUR` on that impl too, so it's not something this PR introduced. Worth handling in the new branch anyway, since it's the difference between `UNIX_TIME` working and not.

All four together:

```suggestion
		// Substrait's UNIX_TIME specifier (epoch seconds, return type i64) has no
		// DuckDB date_part equivalent of that name. Map it to date_part('epoch') and
		// floor before casting, to honor the standard's "elapsed whole seconds".
		if (StringUtil::CIEquals(subfield, "UNIX_TIME")) {
			// The precision_timestamp_tz impl carries a trailing timezone argument.
			// UNIX_TIME is timezone-independent, so drop it instead of emitting an
			// unbindable three-argument date_part.
			if (children.size() > 1) {
				children.resize(1);
			}
			children.insert(children.begin(), make_uniq<ConstantExpression>(Value("epoch")));
			auto call = make_uniq<FunctionExpression>(RemapFunctionName(function_name), std::move(children));
			// date_part('epoch') returns DOUBLE and DuckDB's DOUBLE->BIGINT cast rounds
			// half-to-even, which would report one second too many for any sub-second
			// timestamp. UNIX_TIME is elapsed whole seconds, so floor explicitly.
			vector<unique_ptr<ParsedExpression>> floor_args;
			floor_args.push_back(std::move(call));
			auto floored = make_uniq<FunctionExpression>("floor", std::move(floor_args));
			return make_uniq<CastExpression>(LogicalType::BIGINT, std::move(floored));
		}
```

I built and ran this: all four cases above return the spec value, `unix_time`/`Unix_Time` now resolve, the `req_ptstz_str` plan no longer hits the arity error, and `YEAR`/`MONTH`/`SECOND` are unaffected. SQL suite 1503 assertions / 53 cases and C tests 261 / 50 both stay green. (`StringUtil` is already used in this file, so no new include.)

One last thing, unrelated to the diff but worth knowing: the new code is currently unreachable from the test suite. `test_substrait_extract.test` only drives the *producer*, which can never emit `UNIX_TIME` — `epoch` isn't in `to_substrait.cpp`'s `valid_extract_subfields`. A handcrafted plan through `from_substrait_json` is the only way to cover it, and would have caught items 1 and 4.

thgschwind: Good hunch on the coverage gap — confirmed and closed it. Added test/sql/test_substrait_extract_unix_time.test: since the producer never emits UNIX_TIME, it drives the consumer through from_substrait_json on a handcrafted plan (rewrites a real extract(second) plan's enum to lowercase unix_time, which also pins the case-insensitive match).

And your "would have caught items 1 and 4" was exactly right — I verified it by reverting the fix and rebuilding:
- Item 1: the sub-second rows (.75, 1.5, -0.5) assert floored whole seconds, so the pre-fix nearbyint cast fails at the first assertion.
- Item 4: injecting a trailing string argument reproduces the precision_timestamp_tz 3-arg shape, which hits the binder error without the resize(1) drop.

Both green with the fix in. Thanks again for the thorough review.

andrew-coleman (approved):

PR #273, opened 2026-09-03
fix: emit TIMESTAMPTZ constants as precision_timestamp_tz literals

thgschwind:
The Substrait producer (`DuckDBToSubstrait::TransformConstant`, `src/to_substrait.cpp`) has no case for `LogicalTypeId::TIMESTAMP_TZ`, so any `TIMESTAMPTZ` literal that reaches the producer falls through to the `default:` branch and throws:

```
Not implemented Error: Consuming a value of type TIMESTAMP WITH TIME ZONE is not supported yet
```

### Fix

Add a `TransformTimestampTz` that emits a typed `precision_timestamp_tz` literal — mirroring the existing `TransformTime` (`precision_time`) path — and wire it into the `TransformConstant` switch. The value is the raw microsecond count from the `timestamp_tz_t`, with precision `6` (microseconds), which is exactly what the consumer's `kPrecisionTimestampTz` branch reads back via `ScaleToMicros`.

```cpp
void DuckDBToSubstrait::TransformTimestampTz(const Value &dval, substrait::Expression &sexpr) {
    auto &sval = *sexpr.mutable_literal();
    auto precision_timestamp_tz = sval.mutable_precision_timestamp_tz();
    precision_timestamp_tz->set_precision(6); // microseconds
    precision_timestamp_tz->set_value(dval.GetValueUnsafe<timestamp_tz_t>().value);
}
```

```cpp
case LogicalTypeId::TIMESTAMP_TZ:
    TransformTimestampTz(dval, sexpr);
    break;
```

No other constant path changes; only the previously-unhandled `TIMESTAMP_TZ` id is affected.

### Testing

The regression test in `test/sql/test_substrait_types.test` adds a second row that straddles the filter constant, so the filter stays selective and cannot be folded — forcing the `TIMESTAMPTZ` literal into the plan and exercising the new producer path (with round-trip verification via `PRAGMA enable_verification`).

Full build against DuckDB v1.5.5, both suites green with no regressions:

- **SQL suite** (`test/*`): 1503 assertions / 53 cases pass (1 skipped: `require httpfs`).
- **C unit tests** (`test_substrait_exe`): 261 assertions / 50 cases pass.

### DIFF
--- a/src/to_substrait.cpp
+++ b/src/to_substrait.cpp
@@ -196,6 +196,13 @@ void DuckDBToSubstrait::TransformTimestamp(const Value &dval, substrait::Express
 	sval.set_string(dval.ToString());
 }
 
+void DuckDBToSubstrait::TransformTimestampTz(const Value &dval, substrait::Expression &sexpr) {
+	auto &sval = *sexpr.mutable_literal();
+	auto precision_timestamp_tz = sval.mutable_precision_timestamp_tz();
+	precision_timestamp_tz->set_precision(6); // microseconds
+	precision_timestamp_tz->set_value(dval.GetValueUnsafe<timestamp_tz_t>().value);
+}
+
 void DuckDBToSubstrait::TransformInterval(const Value &dval, substrait::Expression &sexpr) {
 	// Substrait supports two types of INTERVAL (interval_year and interval_day)
 	// whereas DuckDB INTERVAL combines both in one type. Therefore intervals
@@ -275,6 +282,9 @@ void DuckDBToSubstrait::TransformConstant(const Value &dval, substrait::Expressi
 	case LogicalTypeId::TIMESTAMP_NS:
 	case LogicalTypeId::TIMESTAMP:
 		TransformTimestamp(dval, sexpr);
+		break;
+	case LogicalTypeId::TIMESTAMP_TZ:
+		TransformTimestampTz(dval, sexpr);
 		break;
 	case LogicalTypeId::INTERVAL:
 		TransformInterval(dval, sexpr);
--- a/src/to_substrait.cpp
+++ b/src/to_substrait.cpp
@@ -197,10 +197,16 @@ void DuckDBToSubstrait::TransformTimestamp(const Value &dval, substrait::Express
 }
 
 void DuckDBToSubstrait::TransformTimestampTz(const Value &dval, substrait::Expression &sexpr) {
+	auto tz_value = dval.GetValue<timestamp_tz_t>();
+	if (!Value::IsFinite(tz_value)) {
+		// DuckDB encodes +/-infinity as INT64_MAX/MIN micros. Emitting that raw would look
+		// like a real instant near year 294241 to any other consumer, so fail loudly instead.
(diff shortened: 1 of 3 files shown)


src/to_substrait.cpp
```diff
@@ -196,6 +196,13 @@ void DuckDBToSubstrait::TransformTimestamp(const Value &dval, substrait::Express
 	sval.set_string(dval.ToString());
 }
 
+void DuckDBToSubstrait::TransformTimestampTz(const Value &dval, substrait::Expression &sexpr) {
+	auto &sval = *sexpr.mutable_literal();
+	auto precision_timestamp_tz = sval.mutable_precision_timestamp_tz();
+	precision_timestamp_tz->set_precision(6); // microseconds
+	precision_timestamp_tz->set_value(dval.GetValueUnsafe<timestamp_tz_t>().value);
+}
```
andrew-coleman, line 210: Nice catch on using raw micros rather than `ToString()` — a string would render in the session timezone and re-parse in the consumer's, corrupting the instant. One gap though: this silently turns a loud failure into a wrong answer for `±infinity`.

DuckDB stores those as `±INT64_MAX` micros, so they now go out as literals:

```
'infinity'::TIMESTAMPTZ  → {"precisionTimestampTz":{"precision":6,"value":"9223372036854775807"}}
'-infinity'::TIMESTAMPTZ → {"precisionTimestampTz":{"precision":6,"value":"-9223372036854775807"}}
```

Our own consumer round-trips that back to `infinity` only because `INT64_MAX` is DuckDB's private sentinel. substrait-java / DataFusion / Velox / Acero decode it as a real instant near **year 294241**, and any consumer scaling µs→ns overflows int64. Before this PR the producer threw, so the one case Substrait can't represent at least failed loudly.

`Value::IsFinite(timestamp_tz_t)` is already available (`value.hpp:662`) and bottoms out in `Timestamp::IsFinite`, which checks exactly those two sentinels. Switching to the checked `GetValue<>` at the same time also matches `TransformTime` above:

```suggestion
void DuckDBToSubstrait::TransformTimestampTz(const Value &dval, substrait::Expression &sexpr) {
	auto tz_value = dval.GetValue<timestamp_tz_t>();
	if (!Value::IsFinite(tz_value)) {
		// DuckDB encodes +/-infinity as INT64_MAX/MIN micros. Emitting that raw would look
		// like a real instant near year 294241 to any other consumer, so fail loudly instead.
		throw NotImplementedException("Substrait has no representation for infinite TIMESTAMPTZ values");
	}
	auto &sval = *sexpr.mutable_literal();
	auto precision_timestamp_tz = sval.mutable_precision_timestamp_tz();
	precision_timestamp_tz->set_precision(6); // microseconds
	precision_timestamp_tz->set_value(tz_value.value);
}
```

(`GetValueUnsafe` only guards with a `D_ASSERT` that compiles away in release — and `BIGINT`/`TIMESTAMP`/`TIMESTAMP_NS`/`TIME_TZ` are all INT64-backed, so a stray `case` label would reinterpret an unrelated int64 as epoch micros with no diagnostic.)

I verified the suggested version compiles clean against this branch.

andrew-coleman (approved):

PR #274, opened 2026-09-07
fix: honor emit mappings beyond ProjectRel

### LINKED ISSUE #267: `RelCommon.emit` output_mapping is ignored on six relation types, returning the wrong columns
`from_substrait_json` ignores `RelCommon.emit.output_mapping` on ReadRel, FilterRel, SortRel, FetchRel, JoinRel and AggregateRel: the query returns the leading input columns instead of the ones the mapping names, and raises no error, so the caller gets another column's data. ProjectRel is the one relation where the mapping is applied.

### Environment

DuckDB 1.5.5 on macOS arm64, `substrait` extension from the community repository:

```sql
SELECT extension_version, installed_from FROM duckdb_extensions() WHERE extension_name = 'substrait';
```

```
extension_version  installed_from
4ea63dd            community
```

### Reproducing

Every plan below is self-contained: one `virtualTable` row `(c0=10, c1='x', c2=true)` typed `(i64, string, bool)`, all required, so nothing needs to be created first. The mapping is `[2, 0]` throughout, which selects `c2` then `c0`, i.e. expected output `(true, 10)` with types `(BOOLEAN, BIGINT)`.

```sql
LOAD substrait;

-- emit on ReadRel
CALL from_substrait_json('{"relations":[{"root":{"input":{"read":{"common":{"emit":{"outputMapping":[2,0]}},"baseSchema":{"names":["c0","c1","c2"],"struct":{"types":[{"i64":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_REQUIRED"}},{"bool":{"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"i64":"10"}},{"literal":{"string":"x"}},{"literal":{"boolean":true}}]}]}}},"names":["k0","k1"]}}],"version":{"minorNumber":102,"producer":"repro"}}');
```

Observed (column types from the result-set description, then the row):

```
k0 BIGINT, k1 VARCHAR    (10, 'x')
```

The same mapping on the other relations, each over the same one-row virtual table:

| relation carrying `emit [2,0]` | result |
| --- | --- |
| ReadRel | `k0 BIGINT, k1 VARCHAR` → `(10, 'x')` |
| FilterRel | `k0 BIGINT, k1 VARCHAR` → `(10, 'x')` |
| SortRel | `k0 BIGINT, k1 VARCHAR` → `(10, 'x')` |
| FetchRel | `k0 BIGINT, k1 VARCHAR` → `(10, 'x')` |
| JoinRel (inner, `true`) | `k0 BIGINT, k1 VARCHAR` → `(10, 'x')` |
| AggregateRel (group by all three) | `k0 BIGINT, k1 VARCHAR` → `(10, 'x')` |
| ProjectRel | `k0 BOOLEAN, k1 BIGINT` → `(true, 10)` |

<details>
<summary>The other six plans</summary>

```sql
-- FilterRel
CALL from_substrait_json('{"relations":[{"root":{"input":{"filter":{"common":{"emit":{"outputMapping":[2,0]}},"input":{"read":{"baseSchema":{"names":["c0","c1","c2"],"struct":{"types":[{"i64":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_REQUIRED"}},{"bool":{"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"i64":"10"}},{"literal":{"string":"x"}},{"literal":{"boolean":true}}]}]}}},"condition":{"selection":{"directReference":{"structField":{"field":2}},"rootReference":{}}}}},"names":["k0","k1"]}}],"version":{"minorNumber":102,"producer":"repro"}}');

-- SortRel
CALL from_substrait_json('{"relations":[{"root":{"input":{"sort":{"common":{"emit":{"outputMapping":[2,0]}},"input":{"read":{"baseSchema":{"names":["c0","c1","c2"],"struct":{"types":[{"i64":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_REQUIRED"}},{"bool":{"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"i64":"10"}},{"literal":{"string":"x"}},{"literal":{"boolean":true}}]}]}}},"sorts":[{"expr":{"selection":{"directReference":{"structField":{}},"rootReference":{}}},"direction":"SORT_DIRECTION_ASC_NULLS_LAST"}]}},"names":["k0","k1"]}}],"version":{"minorNumber":102,"producer":"repro"}}');

-- FetchRel
CALL from_substrait_json('{"relations":[{"root":{"input":{"fetch":{"common":{"emit":{"outputMapping":[2,0]}},"input":{"read":{"baseSchema":{"names":["c0","c1","c2"],"struct":{"types":[{"i64":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_REQUIRED"}},{"bool":{"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"i64":"10"}},{"literal":{"string":"x"}},{"literal":{"boolean":true}}]}]}}},"countExpr":{"literal":{"i64":"10"}}}},"names":["k0","k1"]}}],"version":{"minorNumber":102,"producer":"repro"}}');

-- JoinRel
CALL from_substrait_json('{"relations":[{"root":{"input":{"join":{"common":{"emit":{"outputMapping":[2,0]}},"left":{"read":{"baseSchema":{"names":["c0","c1","c2"],"struct":{"types":[{"i64":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_REQUIRED"}},{"bool":{"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"i64":"10"}},{"literal":{"string":"x"}},{"literal":{"boolean":true}}]}]}}},"right":{"read":{"baseSchema":{"names":["c0","c1","c2"],"struct":{"types":[{"i64":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_REQUIRED"}},{"bool":{"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"i64":"10"}},{"literal":{"string":"x"}},{"literal":{"boolean":true}}]}]}}},"expression":{"literal":{"boolean":true}},"type":"JOIN_TYPE_INNER"}},"names":["k0","k1"]}}],"version":{"minorNumber":102,"producer":"repro"}}');

-- AggregateRel
CALL from_substrait_json('{"relations":[{"root":{"input":{"aggregate":{"common":{"emit":{"outputMapping":[2,0]}},"input":{"read":{"baseSchema":{"names":["c0","c1","c2"],"struct":{"types":[{"i64":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_REQUIRED"}},{"bool":{"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"i64":"10"}},{"literal":{"string":"x"}},{"literal":{"boolean":true}}]}]}}},"groupings":[{"expressionReferences":[0,1,2]}],"groupingExpressions":[{"selection":{"directReference":{"structField":{}},"rootReference":{}}},{"selection":{"directReference":{"structField":{"field":1}},"rootReference":{}}},{"selection":{"directReference":{"structField":{"field":2}},"rootReference":{}}}]}},"names":["k0","k1"]}}],"version":{"minorNumber":102,"producer":"repro"}}');

-- ProjectRel (this one is correct)
CALL from_substrait_json('{"relations":[{"root":{"input":{"project":{"common":{"emit":{"outputMapping":[2,0]}},"input":{"read":{"baseSchema":{"names":["c0","c1","c2"],"struct":{"types":[{"i64":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_REQUIRED"}},{"bool":{"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"i64":"10"}},{"literal":{"string":"x"}},{"literal":{"boolean":true}}]}]}}},"expressions":[{"selection":{"directReference":{"structField":{"field":1}},"rootReference":{}}}]}},"names":["k0","k1"]}}],"version":{"minorNumber":102,"producer":"repro"}}');
```

</details>

The mapping is not being truncated to fit, it is not being read at all. With `outputMapping [2,1,0]` and three root names, so the same column count in and out, the read still returns the input order:

```sql
CALL from_substrait_json('{"relations":[{"root":{"input":{"read":{"common":{"emit":{"outputMapping":[2,1,0]}},"baseSchema":{"names":["c0","c1","c2"],"struct":{"types":[{"i64":{"nullability":"NULLABILITY_REQUIRED"}},{"string":{"nullability":"NULLABILITY_REQUIRED"}},{"bool":{"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"i64":"10"}},{"literal":{"string":"x"}},{"literal":{"boolean":true}}]}]}}},"names":["k0","k1","k2"]}}],"version":{"minorNumber":102,"producer":"repro"}}');
```

```
k0 BIGINT, k1 VARCHAR, k2 BOOLEAN    (10, 'x', true)
```

expected `(true, 'x', 10)`.

### Where this appears to come from

At `4ea63dd`, `GetOutputMapping` (`src/from_substrait.cpp:1153`) has exactly one caller, `TransformProjectOp` (`src/from_substrait.cpp:1188`). No other `Transform*Op` reads `common().emit()`, so on the other relations the mapping is never consulted.

The column count comes from somewhere else. `TransformRootOp` (`src/from_substrait.cpp:1856`) asks `GetProjection` (`:1841`) for a projection to read column definitions from. For a read, join or aggregate child that call returns null. The `else` branch at `:1876`-`:1879` then builds one `PositionalReferenceExpression(1..n)`, one per name on the root. A well-formed root carries one name per emitted field, so the column count comes out right and only the contents are wrong.

### Uncertain

I did not check the physical relations (`hash_join`, `merge_join`, `nested_loop_join`, `exchange`, `expand`, `window`, `set`, `lateral_join`, `cross`). `GetCommon` covers them, but `GetOutputMapping` has no caller for them either, so I would expect the same. I have not run it.

alexandrefimov:
RelCommon.emit was only applied by ProjectRel. Other relations returned their natural output order, so a mapping such as [2, 0] could silently return the first two columns instead.

Apply the [emit mapping](https://substrait.io/relations/common_fields/#emit) after converting supported query relations, including after ReadRel projection. Use the requested field order and preserve duplicates, and keep ProjectRel's existing mapping path so it is applied once. Add regression coverage for the six reported relation types, Cross, Set, Window and LateralJoin, plus Project composition.

Validate mapping indices and reject explicitly empty mappings, since DuckDB cannot return zero columns. Also reject emit on a right-side FilterRel that the lateral converter extracts into its join condition, where the mapping would otherwise still be lost.

For DELETE, unwrap output projections before reading the filter. Reject inputs whose filter cannot be applied directly to the target table, including a filter over emitted columns. Cover both the remaining rows and rejection without changing the table.

Closes #267

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -975,6 +975,9 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformLateralJoinOp(const substrait::
 		left_column_count = left_rel->Columns().size();
 
 		auto &filter_rel = slateral.right().filter();
+		if (filter_rel.common().has_emit()) {
+			throw NotImplementedException("emit on the right-side FilterRel of a LateralJoinRel is not supported");
+		}
 		auto &filter_condition = filter_rel.condition();
 		// Transform the filter condition; field indices will be relative to the right side
 		correlated_filter_condition = TransformExpr(filter_condition);
@@ -1156,7 +1159,11 @@ const google::protobuf::RepeatedField<int32_t> &GetOutputMapping(const substrait
 		static google::protobuf::RepeatedField<int32_t> empty_mapping;
 		return empty_mapping;
 	}
-	return common->emit().output_mapping();
+	const auto &mapping = common->emit().output_mapping();
+	if (mapping.empty()) {
+		throw NotImplementedException("An empty emit mapping (zero output columns) is not supported");
+	}
+	return mapping;
 }
 
 shared_ptr<Relation>
@@ -1171,7 +1178,7 @@ SubstraitToDuckDB::TransformProjectOp(const substrait::Rel &sop,
 	size_t num_input_columns = 0;
 	if (sop.project().input().rel_type_case() == substrait::Rel::RelTypeCase::kRead) {
 		auto &sget = sop.project().input().read();
-		if (sget.has_virtual_table()) {
+		if (sget.has_virtual_table() && !sget.common().has_emit()) {
 			auto virtual_table = sget.virtual_table();
 			if ((virtual_table.expressions().empty()) ||
 			    (virtual_table.expressions().size() > 0 && virtual_table.expressions(0).fields().empty())) {
@@ -1197,6 +1204,9 @@ SubstraitToDuckDB::TransformProjectOp(const substrait::Rel &sop,
 	} else {
 		expressions.resize(mapping.size());
 		for (size_t i = 0; i < mapping.size(); i++) {
+			if (mapping[i] < 0) {
+				throw InvalidInputException("Project emit mapping index must be non-negative, got %d", mapping[i]);
+			}
 			if (mapping[i] < num_input_columns) {
 				expressions[i] = make_uniq<PositionalReferenceExpression>(mapping[i] + 1);
 			} else {
@@ -1789,29 +1799,41 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformReferenceOp(const substrait::Re
 
 shared_ptr<Relation> SubstraitToDuckDB::TransformOp(const substrait::Rel &sop,
                                                     const google::protobuf::RepeatedPtrField<std::string> *names) {
+	shared_ptr<Relation> result;
 	switch (sop.rel_type_case()) {
 	case substrait::Rel::RelTypeCase::kJoin:
-		return TransformJoinOp(sop);
+		result = TransformJoinOp(sop);
+		break;
 	case substrait::Rel::RelTypeCase::kLateralJoin:
-		return TransformLateralJoinOp(sop);
+		result = TransformLateralJoinOp(sop);
+		break;
 	case substrait::Rel::RelTypeCase::kCross:
-		return TransformCrossProductOp(sop);
+		result = TransformCrossProductOp(sop);
+		break;
 	case substrait::Rel::RelTypeCase::kFetch:
-		return TransformFetchOp(sop, names);
+		result = TransformFetchOp(sop, names);
+		break;
 	case substrait::Rel::RelTypeCase::kFilter:
-		return TransformFilterOp(sop);
+		result = TransformFilterOp(sop);
+		break;
 	case substrait::Rel::RelTypeCase::kProject:
+		// Project applies emit while selecting its input columns and expressions.
 		return TransformProjectOp(sop, names);
 	case substrait::Rel::RelTypeCase::kAggregate:
-		return TransformAggregateOp(sop);
+		result = TransformAggregateOp(sop);
+		break;
 	case substrait::Rel::RelTypeCase::kRead:
-		return TransformReadOp(sop);
+		result = TransformReadOp(sop);
+		break;
 	case substrait::Rel::RelTypeCase::kSort:
-		return TransformSortOp(sop, names);
+		result = TransformSortOp(sop, names);
+		break;
 	case substrait::Rel::RelTypeCase::kWindow:
-		return TransformWindowOp(sop);
+		result = TransformWindowOp(sop);
+		break;
 	case substrait::Rel::RelTypeCase::kSet:
-		return TransformSetOp(sop, names);
+		result = TransformSetOp(sop, names);
+		break;
 	case substrait::Rel::RelTypeCase::kWrite:
 		return TransformWriteOp(sop);
 	case substrait::Rel::RelTypeCase::kReference:
@@ -1820,6 +1842,23 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformOp(const substrait::Rel &sop,
 		throw NotImplementedException("Unsupported relation type %s",
 		                              string(substrait::Rel::GetDescriptor()->FindFieldByNumber(sop.rel_type_case())->name()));
 	}
+
+	const auto &mapping = GetOutputMapping(sop);
+	if (mapping.empty()) {
+		return result;
+	}
+	const auto column_count = result->Columns().size();
+	vector<unique_ptr<ParsedExpression>> expressions;
+	vector<string> aliases;
+	for (auto index : mapping) {
+		if (index < 0 || static_cast<idx_t>(index) >= column_count) {
+			throw InvalidInputException("Relation emit mapping index %d is out of range for %llu output columns", index,
+			                            static_cast<unsigned long long>(column_count));
+		}
+		expressions.push_back(make_uniq<PositionalReferenceExpression>(static_cast<idx_t>(index) + 1));
+		aliases.push_back("expr_" + to_string(aliases.size()));
+	}
+	return make_shared_ptr<ProjectionRelation>(std::move(result), std::move(expressions), std::move(aliases));
 }
 
 void SkipColumnNamesRecurse(int32_t &columns_to_skip, const LogicalType &type) {
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -1768,19 +1768,22 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformWriteOp(const substrait::Rel &s
 	case substrait::WriteRel::WriteOp::WriteRel_WriteOp_WRITE_OP_INSERT:
 		return input->InsertRel(schema_name, table_name);
 	case substrait::WriteRel::WriteOp::WriteRel_WriteOp_WRITE_OP_DELETE: {
-		switch (input->type) {
-		case RelationType::PROJECTION_RELATION: {
-			auto project = std::move(input.get()->Cast<ProjectionRelation>());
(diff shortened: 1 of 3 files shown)


src/from_substrait.cpp
```diff
@@ -1789,29 +1802,41 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformReferenceOp(const substrait::Re
...
 	case substrait::Rel::RelTypeCase::kWindow:
-		return TransformWindowOp(sop);
+		result = TransformWindowOp(sop);
+		break;
 	case substrait::Rel::RelTypeCase::kSet:
-		return TransformSetOp(sop, names);
+		result = TransformSetOp(sop, names);
+		break;
 	case substrait::Rel::RelTypeCase::kWrite:
 		return TransformWriteOp(sop);
```
EpsilonPrime: Nit: WriteRel also carries a `RelCommon`, so an emit there is still silently dropped. It's rare, but in keeping with the lateral-filter rejection, consider throwing NotImplemented when `sop.write().common().has_emit()`.

src/from_substrait.cpp
```diff
@@ -1820,6 +1845,23 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformOp(const substrait::Rel &sop,
...
+	const auto &mapping = GetOutputMapping(sop);
+	if (mapping.empty()) {
+		return result;
+	}
+	const auto column_count = result->Columns().size();
+	vector<unique_ptr<ParsedExpression>> expressions;
+	vector<string> aliases;
+	for (auto index : mapping) {
+		if (index < 0 || static_cast<idx_t>(index) >= column_count) {
+			throw InvalidInputException("Relation emit mapping index %d is out of range for %llu output columns", index,
```
EpsilonPrime: With more than one grouping set, Substrait appends a grouping set index column to an AggregateRel's output, which DuckDB doesn't produce. An external plan whose emit references that column lands here with the generic out-of-range error, which doesn't explain why. Could you give that case its own message? For example, when `sop` is an aggregate with `groupings_size() > 1` and `index == column_count`, throw something like `NotImplementedException("Emit references the grouping set index column, which is not supported")`.

(Related: #291 makes the producer emit `[0..n)` on these aggregates so the index column is left out. Either merge order works.)

src/from_substrait.cpp
alexandrefimov: Done in 55446be.

src/from_substrait.cpp
alexandrefimov: Done in 55446be. An index past that column, or on an aggregate with one grouping set, still gets the out-of-range error.

src/from_substrait.cpp
```diff
@@ -1758,19 +1768,22 @@ shared_ptr<Relation> SubstraitToDuckDB::TransformWriteOp(const substrait::Rel &s
...
-			auto filter = std::move(input.get()->Cast<FilterRelation>());
-			return make_shared_ptr<DeleteRelation>(filter.context, std::move(filter.condition), catalog_name, schema_name, table_name);
+		if (delete_input->type != RelationType::FILTER_RELATION) {
+			throw NotImplementedException("Unsupported relation type for delete operation");
 		}
-		default:
+		auto &filter = delete_input->Cast<FilterRelation>();
+		// DeleteRelation binds the predicate directly against the target table. A predicate
+		// over projected or filtered input cannot be reused without translating its meaning.
+		if (filter.child->type != RelationType::TABLE_RELATION && filter.child->type != RelationType::VIEW_RELATION) {
```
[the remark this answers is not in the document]
alexandrefimov: Fixed in 93f00f2: the filter must read the write target, otherwise the delete is rejected.

PR #275, opened 2026-09-07
fix: preserve timestamp literal precision

### LINKED ISSUE #268: `PRECISION_TIMESTAMP` literals are always converted to microseconds, losing digits at precision 9
A `precision_timestamp` literal in a `ReadRel.VirtualTable` comes back as `TIMESTAMP` (microseconds) whatever precision it declares. At precision 9 the last three digits are dropped. The same `base_schema` on a read with no rows yields `TIMESTAMP_NS`, so the two paths disagree about the same declared type, and they disagree twice: about the carrying type, and about which precisions they accept at all. `ScaleToMicros` (`src/from_substrait.cpp:138-148`) takes 0 through 9, while `SubstraitToDuckType` (`:659-673`) handles only 0, 3, 6 and 9, so a plan declaring precision 7 in `base_schema` is refused with a clear message while the same precision 7 as a literal is accepted and silently truncated.

### Environment

DuckDB 1.5.5 on macOS arm64, `substrait` extension from the community repository:

```sql
SELECT extension_version, installed_from FROM duckdb_extensions() WHERE extension_name = 'substrait';
```

```
extension_version  installed_from
4ea63dd            community
```

### Reproducing

One row, one column, `precisionTimestamp` precision 9, value `1234567891`, i.e. 1.234567891 s after the epoch:

```sql
LOAD substrait;

CALL from_substrait_json('{"relations":[{"root":{"input":{"read":{"baseSchema":{"names":["c"],"struct":{"types":[{"precisionTimestamp":{"precision":9,"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{"expressions":[{"fields":[{"literal":{"precisionTimestamp":{"precision":9,"value":"1234567891"}}}]}]}}},"names":["c"]}}],"version":{"minorNumber":102,"producer":"repro"}}');
```

```
column type: TIMESTAMP
value:       1970-01-01 00:00:01.234567
```

The `891` nanoseconds are gone and nothing is reported.

The same `base_schema` with an empty `virtualTable` (no rows, so only the declared type is used):

```sql
CALL from_substrait_json('{"relations":[{"root":{"input":{"read":{"baseSchema":{"names":["c"],"struct":{"types":[{"precisionTimestamp":{"precision":9,"nullability":"NULLABILITY_REQUIRED"}}],"nullability":"NULLABILITY_REQUIRED"}},"virtualTable":{}}},"names":["c"]}}],"version":{"minorNumber":102,"producer":"repro"}}');
```

```
column type: TIMESTAMP_NS
(0 rows)
```

Same plan shape, same declared type, two different DuckDB types.

Sweeping the precision on the one-row form, literal value `1234567891` throughout:

| declared precision | column type | value |
| --- | --- | --- |
| 3 | `TIMESTAMP` | `1970-01-15 06:56:07.891000` |
| 6 | `TIMESTAMP` | `1970-01-01 00:20:34.567891` |
| 9 | `TIMESTAMP` | `1970-01-01 00:00:01.234567` |

The instant is right in each case (the value is scaled); it is the carrying type that is fixed at microseconds, which only costs data at 9.

### Where this appears to come from

At `4ea63dd`, `TransformLiteralToValue` (`src/from_substrait.cpp:245-251`) scales the value to microseconds and returns `Value::TIMESTAMP`, whatever precision the literal declared. That branch arrived with #247, which added the missing literal types.

`SubstraitToDuckType` (`src/from_substrait.cpp:659-673`) already maps precision 9 to `TIMESTAMP_NS`, but a non-empty virtual table never reaches it: `TransformReadOp` sends `virtual_table().expressions()` to `GetValuesExpression` (`:1393`), which transforms the literals and does not look at `base_schema`. The `SubstraitToDuckType` call at `:1406` is on the empty-table branch, which is why that one comes out as `TIMESTAMP_NS`.

`precision_timestamp_tz` (`:252-258`) does the same thing, though there `SubstraitToDuckType` (`:675-682`) also accepts only precision 6, so the two agree.

### Expected behaviour

A literal declaring precision 9 declares nanoseconds, and the value `1234567891` carries nine sub-second digits. `SubstraitToDuckType` already maps that to `TIMESTAMP_NS`; the literal path should reach the same type.

### Uncertain

I only exercised the literal through `ReadRel.VirtualTable`. `TransformLiteralToValue` is shared, so a `precision_timestamp` literal appearing in a projection or a filter should behave the same way, but I did not run those.

alexandrefimov:
A precision_timestamp literal was always converted to microseconds. At precision 9 this dropped the last three digits and returned TIMESTAMP instead of TIMESTAMP_NS. An empty virtual read with the same declared type already returned TIMESTAMP_NS.

Construct timestamp values in their [declared units](https://github.com/substrait-io/substrait/blob/v0.98.0/proto/substrait/algebra.proto#L1151-L1155) for the supported precisions 0, 3, 6, and 9. Now reject other precisions with the same error as type conversion, including 1, 2, 4, 5, 7, and 8, which were previously accepted and converted to microseconds. Keep precision_timestamp_tz handling unchanged.

Add regression coverage for exact types and values in populated and empty virtual reads, projects, nanosecond-sensitive filters, and nested list literals, including negative values and unsupported precisions.

Closes #268

### DIFF
--- a/src/from_substrait.cpp
+++ b/src/from_substrait.cpp
@@ -245,9 +245,18 @@ Value TransformLiteralToValue(const substrait::Expression_Literal &literal) {
 	case substrait::Expression_Literal::LiteralTypeCase::kPrecisionTimestamp: {
 		int64_t timestamp_value = literal.precision_timestamp().value();
 		int32_t precision = literal.precision_timestamp().precision();
-		int64_t micros = ScaleToMicros(timestamp_value, precision);
-		timestamp_t ts(micros);
-		return Value::TIMESTAMP(ts);
+		switch (precision) {
+		case 0:
+			return Value::TIMESTAMPSEC(timestamp_sec_t(timestamp_value));
+		case 3:
+			return Value::TIMESTAMPMS(timestamp_ms_t(timestamp_value));
+		case 6:
+			return Value::TIMESTAMP(timestamp_t(timestamp_value));
+		case 9:
+			return Value::TIMESTAMPNS(timestamp_ns_t(timestamp_value));
+		default:
+			throw NotImplementedException("Unsupported timestamp precision: %d", precision);
+		}
 	}
 	case substrait::Expression_Literal::LiteralTypeCase::kPrecisionTimestampTz: {
 		int64_t timestamp_value = literal.precision_timestamp_tz().value();
--- a/test/c/CMakeLists.txt
+++ b/test/c/CMakeLists.txt
@@ -19,7 +19,7 @@ find_package(yaml-cpp REQUIRED)
 # definitions the generated function registry was built from.
 find_package(SubstraitExtensions CONFIG REQUIRED)
 
-set(ALL_SOURCES test_substrait_c_api.cpp test_substrait_c_utils.cpp test_projection.cpp test_base_schema_projection.cpp test_root_names.cpp test_dialect_validation.cpp)
+set(ALL_SOURCES test_substrait_c_api.cpp test_substrait_c_utils.cpp test_projection.cpp test_base_schema_projection.cpp test_root_names.cpp test_dialect_validation.cpp test_timestamp_literals.cpp)
 
 
 add_library_unity(test_substrait OBJECT ${ALL_SOURCES})
--- a/test/c/test_timestamp_literals.cpp
+++ b/test/c/test_timestamp_literals.cpp
@@ -1,0 +1,145 @@
+#include "catch.hpp"
+#include "test_helpers.hpp"
+#include "test_substrait_c_utils.hpp"
+#include <nlohmann/json.hpp>
+
+using namespace duckdb;
+
+namespace {
+using TimestampJSON = nlohmann::json;
+
+TimestampJSON TimestampLiteral(int precision, int64_t value) {
+	return {{"literal", {{"precisionTimestamp", {{"precision", precision}, {"value", std::to_string(value)}}}}}};
+}
+
+TimestampJSON TimestampRead(int precision, const duckdb::vector<int64_t> &values) {
+	auto fields = TimestampJSON::array();
+	for (auto value : values) {
+		fields.push_back({{"fields", {TimestampLiteral(precision, value)}}});
+	}
+	return {{"read",
+	         {{"baseSchema",
+	           {{"names", {"value"}},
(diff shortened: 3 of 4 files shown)


src/from_substrait.cpp
```diff
@@ -245,9 +245,18 @@ Value TransformLiteralToValue(const substrait::Expression_Literal &literal) {
...
+			return Value::TIMESTAMPMS(timestamp_ms_t(timestamp_value));
+		case 6:
+			return Value::TIMESTAMP(timestamp_t(timestamp_value));
+		case 9:
+			return Value::TIMESTAMPNS(timestamp_ns_t(timestamp_value));
+		default:
+			throw NotImplementedException("Unsupported timestamp precision: %d", precision);
+		}
 	}
 	case substrait::Expression_Literal::LiteralTypeCase::kPrecisionTimestampTz: {
```
EpsilonPrime: The reasoning behind this PR applies here too. DuckDB only has a microsecond `TIMESTAMPTZ`, so a `precision_timestamp_tz` literal at any other precision is currently scaled to micros and comes back as a different type than the plan declared. The caller either gets wrong data or would have to convert it back.

The TZ type path already rejects this (`SubstraitToDuckType` throws `DuckDB TIMESTAMP_TZ only supports microsecond precision (6), got: %d`). Could you make the literal path match: reject `precision != 6` with the same error, then build `timestamp_tz_t` directly from the value instead of calling `ScaleToMicros`? A test alongside the existing unsupported-precision cases (e.g. TZ precision 3 and 9 both rejected, 6 accepted) would cover it.

EpsilonPrime (changes requested):
The `precision_timestamp` fix looks right and the tests are thorough. One change needed before approval, noted inline: `precision_timestamp_tz` literals have the same problem this PR fixes for `precision_timestamp`.

src/from_substrait.cpp
alexandrefimov: Done in 5bfd5d2: precision other than 6 now fails with the type path's error, 6 is read as is. Tests cover 3, 9 and 6. The precision 3 tz case in `test_literals.test` now expects the error.

PR #280, opened 2026-09-14
fix: emit integer literals for TINYINT and unsigned integer types

### LINKED ISSUE #279: Producer cannot emit integer literals for TINYINT and unsigned integer types
# Producer cannot emit integer literals for TINYINT and unsigned integer types

## Problem

`get_substrait` fails to convert a query to Substrait when an integer
**literal** of type `TINYINT`, `UTINYINT`, `USMALLINT`, `UINTEGER`, or `UBIGINT`
remains in the logical plan. The producer's constant transform
(`DuckDBToSubstrait::TransformConstant`) has no case for these types and fails
with:

```
Not Implemented Error: Consuming a value of type TINYINT is not supported yet
```

(and likewise for the unsigned types).

The **column/type** mapping for these types already works — a plain `SELECT`
over such columns round-trips fine. Only the **literal** path is missing.

## Reproduction (independent of any optimizer setting)

A literal is normally only exposed to the producer when it is not folded away.
A *selective* filter (rows straddling the constant) cannot be folded by
statistics propagation, so the literal survives even with the full optimizer
enabled. On current `main`:

```sql
CREATE TABLE tinys (t TINYINT);
INSERT INTO tinys VALUES (1), (2);
-- row 1 excluded, row 2 included -> filter is selective, literal survives
CALL get_substrait('SELECT t FROM tinys WHERE t > 1');
-- Not Implemented Error: Consuming a value of type TINYINT is not supported yet
```

The same failure occurs for `UTINYINT`, `USMALLINT`, `UINTEGER`, and `UBIGINT`.

## Scope

`UUID` (and other types such as `TIMETZ`) are also missing from the constant
transform, but they have separate, prerequisite round-trip limitations and are
out of scope for this issue.

thgschwind:
The producer's `TransformConstant` had no case for `TINYINT`, `UTINYINT`,
`USMALLINT`, `UINTEGER`, or `UBIGINT`. Any such literal that survived
optimization (e.g. a selective filter that cannot be folded away) fell through
to the default branch and threw `NotImplementedException: Consuming a value of
type <T> is not supported yet`.

This adds the matching producer paths:

- `TINYINT` -> `i8` (new `TransformTinyInt` helper).
- Unsigned types are upcast to the next-wider signed literal, mirroring
  `DuckToSubstraitType`'s existing type mapping: `UTINYINT`->i16,
  `USMALLINT`->i32, `UINTEGER`->i64, `UBIGINT`->decimal(38,0). Substrait has no
  native unsigned integer types.
- `TransformHugeInt` now uses a checked cast (`GetValue<hugeint_t>()`) so it
  also handles `UBIGINT`, which DuckDB stores as `uint64` rather than `hugeint`.

`UUID` (and other types such as `TIMETZ`) remain missing from the constant
transform on purpose: they have separate, prerequisite round-trip limitations
and are out of scope here.

## Test plan

- [x] New `test/sql/test_substrait_integer_literals.test` covers all five
  types. Each filter is selective (rows straddle the constant) so the literal
  cannot be folded away and must be emitted by the producer; under
  `PRAGMA enable_verification` every `get_substrait` round-trips and compares
  results.
- [x] `test_substrait_types.test` and `test_literals.test` still pass (no
  regressions).

Fixes #279

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Substrait conversion for signed and unsigned integer literals across all supported widths.
  - Added safe handling for large unsigned values, including values represented as decimal types.
  - Corrected `TINYINT` literals to emit as Substrait `i8` values.

- **Tests**
  - Added coverage for signed and unsigned integer literals, including boundary and large-value cases.


test/sql/test_substrait_integer_literals.test
```diff
@@ -0,0 +1,63 @@
...
+CALL get_substrait('SELECT i FROM uints WHERE i > 100');
+
+statement ok
+CREATE TABLE ubigs (b UBIGINT);
+
+statement ok
+INSERT INTO ubigs VALUES (1), (18446744073709551615);
+
+statement ok
+CALL get_substrait('SELECT b FROM ubigs WHERE b > 100');
```
andrew-coleman: The constant `100` fits in `i8`, so none of these four tests can detect a wrong upcast width — the exact regression this PR exists to prevent. I mutation-tested it on the built branch: routing `UTINYINT→TransformTinyInt`, `USMALLINT→TransformSmallInt`, `UINTEGER→TransformInteger`, `UBIGINT→TransformBigInt` (each one width too narrow) leaves **all four tests passing**. `PRAGMA enable_verification` can't help here — `ResultEquals` compares result values, not literal widths.

Constants that exceed the narrower type's max close the gap. I verified these keep the filter in the plan, pass against the current code, and fail against all four mutations above:

```suggestion
CALL get_substrait('SELECT t FROM utinys WHERE t > 150');

statement ok
CREATE TABLE usmalls (s USMALLINT);

statement ok
INSERT INTO usmalls VALUES (1), (40000);

statement ok
CALL get_substrait('SELECT s FROM usmalls WHERE s > 33000');

statement ok
CREATE TABLE uints (i UINTEGER);

statement ok
INSERT INTO uints VALUES (1), (4000000000);

statement ok
CALL get_substrait('SELECT i FROM uints WHERE i > 3000000000');

statement ok
CREATE TABLE ubigs (b UBIGINT);

statement ok
INSERT INTO ubigs VALUES (1), (18446744073709551615);

statement ok
CALL get_substrait('SELECT b FROM ubigs WHERE b > 10000000000000000000');
```

The two `TINYINT` cases at lines 24 and 27 have the same gap (`-1` and `1` fit every width), though there's no narrower type to mis-route to, so they matter less.

Unrelated to this file, but since `UBIGINT` now works: the `mode skip` comment at `test_substrait_types.test:160` ("not supported because TIMETZ comparisons generate a UBIGINT") is stale — that block passes if unskipped. Worth updating or enabling while you're here.

thgschwind: Thanks for your comments - indeed. Looking at it right now...

andrew-coleman (approved):

PR #284, opened 2026-09-17
fix: dialect has missing variants for arithmetic aggregates

andrew-coleman:
In the dialect yaml, the functions sum, avg, min, max, product and mode were missing some of the numeric variants

## Summary by CodeRabbit

- **New Features**
  - Expanded arithmetic aggregate function support for 8-bit and 16-bit integers, 32-bit integers, and single-precision floating-point values.
  - The `sum`, `avg`, `min`, `max`, `product`, and `mode` functions now accept these additional numeric types.


EpsilonPrime (approved):

EpsilonPrime:
median has the same gap.  Substrait Core declares i8 through fp64, but the dialect only lists req_i64 and req_fp64.

PR #286, opened 2026-09-17
fix: emit NOT LIKE as not(like(..)) instead of not_equal

### LINKED ISSUE #282: `NOT LIKE` is emitted as Substrait `not_equal`, causing silent wrong results on round-trip
The producer maps DuckDB's `!~~` (`NOT LIKE`) operator to Substrait `not_equal`, and the consumer compensates by rewriting *any* `not_equal` with a VARCHAR constant back into `!~~`. The two hacks cancel out for some inputs, but they corrupt others — including plain `!=` comparisons that never involved `LIKE` at all.

This produces **wrong query results with no error**, on ordinary SQL, without any external consumer or hand-written plan.

### Reproducer

```sql
CREATE TABLE t(s VARCHAR);
INSERT INTO t VALUES ('a%c'), ('abc');

-- direct SQL
SELECT s FROM t WHERE s != 'a%c';
-- → abc                                  ✅ correct

-- same query, round-tripped through Substrait
SET VARIABLE p = (SELECT json FROM get_substrait_json('SELECT s FROM t WHERE s != ''a%c'''));
SELECT * FROM from_substrait_json(getvariable('p'));
-- → (no rows)                            ❌ wrong
```

`'abc'` is excluded because the `%` in the comparison string is treated as a wildcard: the consumer turned `s != 'a%c'` into `s NOT LIKE 'a%c'`.

andrew-coleman:
The producer mapped DuckDB's `!~~` operator onto Substrait `not_equal`, so a NOT LIKE predicate left this extension as a string inequality. The consumer compensated by rewriting any `not_equal` whose operand was a VARCHAR constant back into `!~~`. The two hacks cancelled for round trips through this extension but corrupted genuine inequality: `WHERE s != 'a%c'` came back as `WHERE s NOT LIKE 'a%c'`, dropping every row that matched the pattern.

NOT LIKE is now emitted as the composition not(like(x, pattern)) -- both are standard functions already declared in duckdb_dialect.yaml -- and `not_equal` is consumed unconditionally as inequality.

Plans previously produced by this extension encoded NOT LIKE as `not_equal` and will now be read as inequality. Those plans were already wrong per the spec, so this is a correctness fix rather than a regression.

Fixes #282

Assisted by AI

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected Substrait conversion so `!=` consistently remains an inequality comparison, including when used with string patterns.
  - Fixed `NOT LIKE` handling to preserve pattern-matching semantics during conversion and round trips.
  - Confirmed that standard `LIKE` behavior remains unchanged.


test/sql/test_substrait_not_like.test
```diff
@@ -0,0 +1,67 @@
...
+statement ok
+PRAGMA enable_verification
+
+statement ok
+CREATE TABLE t (s VARCHAR);
+
+# 'a%c' is both a literal value in the table and a LIKE pattern that matches 'abc'.
+# That is what makes the two operators distinguishable.
+statement ok
+INSERT INTO t VALUES ('a%c'), ('abc'), ('aXbYc'), ('zzz');
```
EpsilonPrime: Could you add a NULL row to `t`? Both `!=` and `NOT LIKE` should drop it, and it'd confirm NULL handling survives the round trip through `not(like(..))`.

src/to_substrait.cpp
```diff
@@ -392,6 +391,27 @@ void DuckDBToSubstrait::TransformFunctionExpression(Expression &dexpr, substrait
...
+		}
+		like_fun->set_function_reference(RegisterFunction("like", like_arg_types));
+		*like_fun->mutable_output_type() = DuckToSubstraitType(LogicalType::BOOLEAN);
+
+		vector<substrait::Type> not_arg_types;
+		not_arg_types.emplace_back(DuckToSubstraitType(LogicalType::BOOLEAN));
+		not_fun->set_function_reference(RegisterFunction("not", not_arg_types));
+		*not_fun->mutable_output_type() = DuckToSubstraitType(dfun.return_type);
+		return;
+	}
```
EpsilonPrime: Nit: this builds the same `not` wrapper as `TransformNotExpression`. That one takes a `BoundOperatorExpression` so it can't be reused directly, but a small shared helper would remove the duplication. Fine to leave as is.

andrew-coleman: I tried this and it did remove 6 lines of duplication, but overall increased the number of lines of code and I think made it slightly less readable. Would prefer to leave as is.

